PR #22, opened 2017-06-27
fix 21 by adding kafka to pom 

wselwood:
so dependencies end up in the nar correctly

Marked as provided in the geomasa-kafka-datastore packages.

### DIFF
--- a/geomesa-nifi-processors/pom.xml
+++ b/geomesa-nifi-processors/pom.xml
@@ -60,6 +60,11 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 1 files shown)


wselwood:
This change I needed because in the nifi nar I was getting a class not found exception when it was trying to connect. With this change it went away. Think it is caused by the kafka dependency in the geomesa-kafka-datastore module being marked as provided, so it doesn't get dragged in to the nar by default. It may work if something else is already making the kafka dependency turn up on the class path.

PR #24, opened 2017-07-14
Kerberos

jrs53:
Updating Hadoop version to fix TGT renewal with Kerberos

### DIFF
--- a/geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/PutGeoMesa.scala
+++ b/geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/PutGeoMesa.scala
@@ -1,5 +1,6 @@
 /***********************************************************************
  * Copyright (c) 2015-2017 Commonwealth Computer Research, Inc.
+ * Portions Crown Copyright (c) 2017 Dstl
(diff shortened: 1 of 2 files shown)


jahhulbert-ccri:
@jrs53 does Kerberos not work with less than 2.7.3 ?

jahhulbert-ccri:
I like 2.7.3 lets go with it

jrs53:
There were a few changes to add the keytab path as a parameter for the Processor, plus fixing param validation. Later, I discovered the requirement to periodically try to renew the Kerberos TGT and while that code works fine outside of NiFi with Hadoop 2.6.0, it didn't work properly inside NiFi hence the new version. Afraid I didn't test anything other than 2.7.3.

jahhulbert-ccri:
weird. the PR files only shows the one pom change...

https://github.com/geomesa/geomesa-nifi/pull/24/files

I'll try to do a command line merge squash to look before merging.

jrs53:
@jnh5y merged the first set of changes already, so the only one left is the Hadoop version. Guess I should have created a separate PR rather than push to my branch - apologies

jahhulbert-ccri:
ah :) no problem!

PR #25, opened 2017-07-17
add custom sft and converter validators

wselwood:
Resolve the two todo comments by adding custom validators for Simple feature types and converters.

Had a lot of problems which boiled down to invalid SFTs so this seemed like the easiest way to fix them. The resulting exceptions are not exactly helpful some of the time when you get this wrong. (Array index out of bounds etc)

### DIFF
--- a/geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/ConverterValidator.scala
+++ b/geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/ConverterValidator.scala
@@ -6,16 +6,11 @@ import org.locationtech.geomesa.convert.{ConfArgs, ConverterConfigResolver}
 /**
   * Simple validator of a convert spec. Tries to parse it to make sure that it can.
   */
-class ConverterValidator extends Validator{
+object ConverterValidator extends Validator{
   override def validate(subject: String, input: String, context: ValidationContext): ValidationResult = {
     ConverterConfigResolver.getArg(ConfArgs(input)) match {
-      case Left(_) => new ValidationResult.Builder().subject(subject).explanation(String.format("'%s' is not a valid converter", subject)).input(input).build;
+      case Left(_) => new ValidationResult.Builder().subject(subject).explanation(s"'$subject' is not a valid converter").input(input).build;
       case Right(_) => new ValidationResult.Builder().subject(subject).input(input).valid(true).build;
     }
   }
 }
-
-object ConverterValidator {
-  def apply() : ConverterValidator = new ConverterValidator()
-}
-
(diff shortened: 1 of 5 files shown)


geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/ConverterValidator.scala
```diff
@@ -0,0 +1,21 @@
+package org.geomesa.nifi.geo.validators
+
+import org.apache.nifi.components.{ValidationContext, ValidationResult, Validator}
+import org.locationtech.geomesa.convert.{ConfArgs, ConverterConfigResolver}
+
+/**
+  * Simple validator of a convert spec. Tries to parse it to make sure that it can.
+  */
+class ConverterValidator extends Validator{
```
elahrvivaz: you could make this implementation an object - it should be thread safe to re-use

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/ConverterValidator.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.apache.nifi.components.{ValidationContext, ValidationResult, Validator}
+import org.locationtech.geomesa.convert.{ConfArgs, ConverterConfigResolver}
+
+/**
+  * Simple validator of a convert spec. Tries to parse it to make sure that it can.
+  */
+class ConverterValidator extends Validator{
+  override def validate(subject: String, input: String, context: ValidationContext): ValidationResult = {
+    ConverterConfigResolver.getArg(ConfArgs(input)) match {
+      case Left(_) => new ValidationResult.Builder().subject(subject).explanation(String.format("'%s' is not a valid converter", subject)).input(input).build;
```
elahrvivaz: [minor] can use scala string interpolation instead of string.format: `s"'$subject' is not a valid converter"`

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/SimpleFeatureTypeValidator.scala
```diff
@@ -0,0 +1,21 @@
+package org.geomesa.nifi.geo.validators
+
+import org.apache.nifi.components.{ValidationContext, ValidationResult, Validator}
+import org.locationtech.geomesa.utils.geotools.{SftArgResolver, SftArgs}
+
+/**
+  * Simple validator that tries to parse a given simple feature type to check that it is valid.
+  */
+class SimpleFeatureTypeValidator extends Validator {
```
elahrvivaz: could make same suggested changes as to the ConverterValidator

PR #28, opened 2017-08-02
Fix #26, #27 Kafka 09,10 nars + HBase ingest

### LINKED ISSUE #26: Kafka 09 and 10 nars
Make nars for each like nifi does

jahhulbert-ccri:


README.md
```diff
@@ -1,44 +1,69 @@
...
+https://github.com/locationtech/geomesa/tree/master/geomesa-tools/conf/sfts
+
+## Defining custom SFTs and Converters
 
-To install the GeoMesa processors you will need to copy the geomesa-nifi-nar file from
-``geomesa-nifi/geomesa-nifi-nar/target/geomesa-nifi-nar-$VERSION.nar`` into the ``lib/`` directory of your NiFi installation.
+There are two ways of providing custom SFTs and converters:
 
-### Install the SFTs and Converters
+* Providing a ``reference.conf`` file via a jar in the ``lib`` dir
```
wselwood: Is it worth noting here that due to the way nifi's classpath works that it can not be in a nar?

jahhulbert-ccri: @wselwood do you mean:
1) a reference.conf file can't be in a nar
2)  a jar containing a reference.conf file can't be in a nar?

wselwood: Both

It can be in the geomesa-nifi nar but not another nar. Due to the way the class loaders work inside nifi if it is in a different nar this one won't be able to see it. (I may have spent a day working this one out)

jahhulbert-ccri: Right. You can embed it in the actual geomesa-nifi nar that you are deploying but not in a separate nar. You can do it in a jar file in the lib dir but this causes some issues. I think I can do some GeoMesa work to the loading of them to make this work as a separate jar properly. I'll make a ticket and let you know.

PR #41, opened 2018-10-12
GEOMESA-2434 Nifi - close converters after use

elahrvivaz:


geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/ConvertToGeoAvro.scala
```diff
@@ -48,20 +51,25 @@ class ConvertToGeoAvro extends AbstractProcessor {
...
   @OnScheduled
   def initialize(context: ProcessContext): Unit = {
     val sft = getSft(context)
     converter = getConverter(sft, context)
-
     getLogger.info(s"Initialized SimpleFeatureConverter sft and converter for type ${sft.getTypeName}")
   }
 
+  @OnRemoved
+  def cleanup(): Unit = {
```
jahhulbert-ccri: i think this should handle things fine.

PR #47, opened 2019-03-19
Permit SFT and Converter to validate

khobbs-ccri:


geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -415,8 +414,7 @@ object AbstractGeoIngestProcessor {
           .name("ConverterName")
           .required(false)
           .description("Choose an SimpleFeature Converter defined by a GeoMesa SFT Provider (preferred)")
-          .allowableValues(ConverterConfigLoader.listConverterNames.sorted: _*)
```
elahrvivaz: I assumed `allowableValues` would perform validation but i guess not... is there any reason to remove it though? I think it makes it a drop down instead of free text (may be wrong there)

khobbs-ccri: It did perform validation, but expressions like ${THE_SFT}, which are not in the list of allowed values are considered invalid. The presence of ".expressionLanguageSupported(..." seems to have overridden it: the drop-down is no-longer there. See khobbs:8080/nifi

elahrvivaz: is that nifi running with these code changes or without them?

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/validators/KnownNameOrExpressionValidator.scala
```diff
@@ -0,0 +1,25 @@
+package org.geomesa.nifi.geo.validators
+
+import org.apache.nifi.components.{ValidationContext, ValidationResult, Validator}
+
+class KnownNameOrExpressionValidator(names: List[String]) extends Validator{
```
elahrvivaz: this could use a little formatting cleanup:

- [ ] space before the `{` after `extends Validator`
- [ ] the `validate` method looks like it could be a single line, if not then indent like [this](https://github.com/locationtech/geomesa/blob/master/geomesa-jobs/src/main/scala/org/locationtech/geomesa/jobs/mapreduce/ConverterInputFormat.scala#L91-L94)
- [ ] add `{}` around the validate function body, since it's not a single line
- [ ] add `{}` around your if/else clauses
- [ ] add new line at the end of the file

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
elahrvivaz: It appears that allowable values are not compatible with expression language support: https://issues.apache.org/jira/browse/NIFI-1832
I'm not crazy about not having it be a drop-down - I think that is going to be a pain for most people. @jahhulbert-ccri @jnh5y what do you guys think?

jnh5y: I think having the drop down would be preferable.  Are there any other options here?  E.g., either provide a nifiExpressionConverterName or selecting something from the drop down?

jahhulbert-ccri: hmm i assumed drop down is preferable too from an easy of use perspective...I guess I could reconsider but my default position are that the users of it would like to see the list of things available...We could always change the OTHER sft name and convertername to use expression language and pull it from there?

jnh5y: @jahhulbert-ccri I like it.  Does that make sense as an option to everyone else?

elahrvivaz: which fields are you referring to? we have feature name override, but that by definition does not match any of the dropdown names. And there is no other converter name field. Are you suggesting adding two new fields?

elahrvivaz:
@khobbs-ccri @jnh5y I think we need to re-evaluate this - currently, the PR turns the sft/converter fields from drop-downs to free text, which IMO we should avoid. Not sure what a better solution would be - additional field for free text/prop replacement?

elahrvivaz:
going to close this for now, shout if we need to still address the issue.

PR #49, opened 2019-05-08
dropwizard latency processor

amcdonaldccri:
- todo: test on the lowside

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <accumulo.version>1.9.2</accumulo.version>
         <hbase.version>1.3.1</hbase.version>
         <kafka.version>1.0.1</kafka.version>
-        <nifi.version>1.8.0</nifi.version>
+        <nifi.version>1.9.2</nifi.version>
         <hadoop.version>2.7.4</hadoop.version>
 
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -256,6 +256,12 @@
                 <version>1.1.1.6</version>
             </dependency>
 
+            <dependency>
+                <groupId>io.dropwizard.metrics</groupId>
+                <artifactId>metrics-graphite</artifactId>
+                <version>4.0.5</version>
```
elahrvivaz: Our current metrics version in geomesa is set to 3.1.2. We were discussing updating it. Is this going to cause compatibility issues? cc @jahhulbert-ccri

elahrvivaz:
starting with 2.4.0, we expose a [metrics registry](https://github.com/locationtech/geomesa/blob/master/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert2/metrics/ConverterMetrics.scala) directly in the [evaluation context](https://github.com/locationtech/geomesa/blob/master/geomesa-convert/geomesa-convert-common/src/main/scala/org/locationtech/geomesa/convert/EvaluationContext.scala#L40). We also allow for configuring the metrics connection via typesafe config. We could probably make this PR a lot simpler by leveraging those.
Is it desirable to have this as a separate processor, or would it be better to integrate it into the ingest/convert processors? That would avoid parsing the input twice, but maybe it's better to keep separate.
Another thought is that we could potentially have the dtg validator do this, either by default or as an option, and not tie it to nifi.

elahrvivaz:
https://geomesa.atlassian.net/browse/GEOMESA-2647

PR #51, opened 2019-09-23
GEOMESA-2712 Fix classpath for loading config files from s3

elahrvivaz:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,12 @@
         <hbase.version>1.3.1</hbase.version>
         <kafka.version>1.0.1</kafka.version>
         <nifi.version>1.8.0</nifi.version>
-        <hadoop.version>2.7.4</hadoop.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -72,7 +72,12 @@
         <hbase.version>1.3.1</hbase.version>
         <kafka.version>1.0.1</kafka.version>
         <nifi.version>1.8.0</nifi.version>
-        <hadoop.version>2.7.4</hadoop.version>
+        <hadoop.version>2.8.5</hadoop.version>
+        <aws.sdk.version>1.11.636</aws.sdk.version>
```
jnh5y: Hmm... seems like including AWS is both a Good Thing and a thing which may cause us trouble...

elahrvivaz: we need it to read s3a:// urls through hadoop - what trouble would it cause?

PR #53, opened 2019-09-25
GEOMESA-2715 Support for caching feature writers

elahrvivaz:


geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -51,14 +53,8 @@ abstract class AbstractGeoIngestProcessor(dataStoreProperties: Seq[PropertyDescr
   private var descriptors: Seq[PropertyDescriptor] = _
   private var relationships: Set[Relationship] = _
 
-  @volatile
-  private var dataStore: DataStore = _
```
jnh5y: I wonder if the DataStore was volatile for some reason?

elahrvivaz: it's wrapped in the Writer now, which is still volatile, so I think it should be equivalent. not sure why it's volatile to begin with - usually that has to do with serialization.

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -169,17 +167,20 @@ abstract class AbstractGeoIngestProcessor(dataStoreProperties: Seq[PropertyDescr
   }
 
   @OnRemoved
-  @OnDisabled
+  @OnStopped
+  @OnShutdown
```
jnh5y: Do we want this to run OnDisabled?  Who else has read through these annotations before?

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -169,17 +167,20 @@ abstract class AbstractGeoIngestProcessor(dataStoreProperties: Seq[PropertyDescr
...
   def cleanup(): Unit = {
-    if (converterPool != null) {
-      CloseWithLogging(converterPool)
-      converterPool = null
```
jnh5y: Is there any converter pooling still going on?

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
elahrvivaz: OnDisabled only applies to ControllerServices: https://github.com/apache/nifi/blob/master/nifi-api/src/main/java/org/apache/nifi/annotation/lifecycle/OnDisabled.java#L47-L48

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -212,82 +211,82 @@ abstract class AbstractGeoIngestProcessor(dataStoreProperties: Seq[PropertyDescr
       if (value == null) { Seq.empty } else {
         Seq(p.getName -> value)
       }
-    }.toMap
-    val props = configured ++ static // note: static props take precedence
-    logger.trace(s"DataStore properties: $props")
+    }
+    val props = configured.toMap ++ static // note: static props take precedence
```
jnh5y: This is just a style change, right?  (e.g. having the .toMap on a line rather than dangling around?)

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
```diff
@@ -212,82 +211,82 @@ abstract class AbstractGeoIngestProcessor(dataStoreProperties: Seq[PropertyDescr
...
+    logger.trace(s"DataStore properties: ${props.map { case (k, v) => s"$k -> $v" }.mkString(", ")}")
```
jnh5y: Idly, I wonder what the toString of a Scala map is...  Would it be much different than this?

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
elahrvivaz: yeah, it's just down on line 419: https://github.com/geomesa/geomesa-nifi/pull/53/files#diff-460fdeb97359e630d090e2f325a18dc5R419

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
elahrvivaz: yeah

geomesa-nifi-processors/src/main/scala/org/geomesa/nifi/geo/AbstractGeoIngestProcessor.scala
elahrvivaz: it just includes `Map(foo -> bar, bar -> baz)` instead of `foo -> bar, bar -> baz`

PR #58, opened 2019-10-24
GEOMESA-2732 Fix dispose of shared controller service data store

elahrvivaz:


geomesa-nifi-processors-accumulo/src/main/scala/org/geomesa/nifi/accumulo/PutGeoMesaAccumulo.scala
```diff
@@ -34,6 +34,8 @@ class PutGeoMesaAccumulo extends AbstractGeoIngestProcessor(PutGeoMesaAccumulo.A
   @volatile
   protected var useControllerService: Boolean = false
 
+  override protected def useDataStoreControllerService: Boolean = useControllerService
```
jnh5y: Ok, why the 2.3 branch different?

elahrvivaz: the code around loading the data store changed in 2.3. the data store itself isn't accessible in the subclass anymore, so i added an override hook. i did the 2.1 version first, otherwise i would have made them consistent...

PR #62, opened 2019-11-13
Updating nifi version to 1.10.0

elahrvivaz:


elahrvivaz:
@jahhulbert let me know if this is going to be a problem, we can add a profile and/or release 2 different versions. If the maven artifacts are built with 1.10.0, but on github we host nars for 1.8 and 1.10, would that work?

PR #72, opened 2020-05-18
GEOMESA-2853 Configure converters through flowfile attributes

elahrvivaz:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,7 @@
 
         <specs2.version>4.3.2</specs2.version>
         <junit.version>4.12</junit.version>
+        <maven.test.jvmargs>-Xms512m -Xmx4g -XX:-UseGCOverheadLimit</maven.test.jvmargs>
 
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
@@ -556,6 +557,37 @@
                         <source>1.8</source>
                         <target>1.8</target>
                         <scalaVersion>${scala.version}</scalaVersion>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <!-- 3.0.0-M4 was released 2019-11-13 -->
+                    <version>3.0.0-M4</version>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.apache.maven.surefire</groupId>
+                            <artifactId>surefire-junit4</artifactId>
+                            <version>3.0.0-M4</version>
+                        </dependency>
+                    </dependencies>
+                    <configuration>
+                        <includes>
+                            <include>**/*Test.java</include>
+                        </includes>
+                        <argLine>${maven.test.jvmargs}</argLine>
+                        <systemPropertyVariables>
+                            <java.net.preferIPv4Stack>true</java.net.preferIPv4Stack>
+                            <java.awt.headless>true</java.awt.headless>
+                            <user.timezone>UTC</user.timezone>
+                            <user.language>en</user.language>
+                            <user.region>US</user.region>
+                            <geomesa.scan.ranges.target>500</geomesa.scan.ranges.target>
+                            <geomesa.accumulo.test.tablet.servers>1</geomesa.accumulo.test.tablet.servers>
+                            <geomesa.hbase.test.region.servers>1</geomesa.hbase.test.region.servers>
+                        </systemPropertyVariables>
+                        <!-- these can interfere with the mini clusters we spin up for unit tests -->
+                        <excludedEnvironmentVariables>HADOOP_HOME,ZOOKEEPER_HOME,ACCUMULO_HOME,HBASE_HOME</excludedEnvironmentVariables>
                     </configuration>
                 </plugin>
                 <plugin>
--- a/geomesa-accumulo-bundle/geomesa-accumulo-processors/pom.xml
+++ b/geomesa-accumulo-bundle/geomesa-accumulo-processors/pom.xml
@@ -18,6 +18,12 @@
         <dependency>
             <groupId>org.geomesa.nifi</groupId>
             <artifactId>geomesa-datastore-processors</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-beanutils</groupId>
+                    <artifactId>commons-beanutils-core</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
@@ -57,6 +63,38 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <version>2.8.47</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-test</artifactId>
+            <version>${accumulo.version}</version>
+            <scope>test</scope>
+            <exclusions>
+                <!-- odd issue resolving some apacheds dependencies -->
+                <exclusion>
+                    <groupId>org.apache.directory.server</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-accumulo-datastore_${scala.binary.version}</artifactId>
+            <version>${geomesa.version}</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-minicluster</artifactId>
+            <version>${accumulo.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>jline</groupId>
+            <artifactId>jline</artifactId>
+            <version>2.11</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/geomesa-accumulo-bundle/geomesa-accumulo-processors/src/test/resources/application.conf
+++ b/geomesa-accumulo-bundle/geomesa-accumulo-processors/src/test/resources/application.conf
@@ -1,0 +1,32 @@
+geomesa = {
+  sfts = {
+    "example" = {
+      attributes = [
+        { name = "fid",  type = "Integer" }
+        { name = "name", type = "String"  }
+        { name = "age",  type = "Integer" }
+        { name = "dtg",  type = "Date",   default = true }
+        { name = "geom", type = "Point",  srid = 4326, default = true }
+      ]
+    }
+  }
+  converters = {
+    "example-csv" = {
+      type   = "delimited-text"
+      format = "CSV"
+      options {
+        skip-lines = 1
+      }
+      id-field = "toString($fid)",
+      fields = [
+        { name = "fid",  transform = "$1::int"                }
+        { name = "name", transform = "$2::string"             }
+        { name = "age",  transform = "$3::int"                }
+        { name = "dtg",  transform = "date('yyyy-MM-dd', $4)" }
+        { name = "lon",  transform = "$7::double"             }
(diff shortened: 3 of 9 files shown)


geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/resources/log4j.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jnh5y, line 1: Hmm... how many log4j.xml files do we have:)?  That can be a later thing (obviously).

elahrvivaz: ;)

pom.xml
```diff
@@ -474,6 +475,37 @@
...
+                            <java.awt.headless>true</java.awt.headless>
+                            <user.timezone>UTC</user.timezone>
+                            <user.language>en</user.language>
+                            <user.region>US</user.region>
+                            <geomesa.scan.ranges.target>500</geomesa.scan.ranges.target>
+                            <geomesa.accumulo.test.tablet.servers>1</geomesa.accumulo.test.tablet.servers>
+                            <geomesa.hbase.test.region.servers>1</geomesa.hbase.test.region.servers>
+                        </systemPropertyVariables>
+                        <!-- these can interfere with the mini clusters we spin up for unit tests -->
+                        <excludedEnvironmentVariables>HADOOP_HOME,ZOOKEEPER_HOME,ACCUMULO_HOME,HBASE_HOME</excludedEnvironmentVariables>
```
jnh5y, line 506: Not something for now, but it would be somewhat amusing if we found a good way to provide configuration to point a unit test suite at running cluster.  *shrugs*

elahrvivaz: that wouldn't be very complicated tbh. i've done it before by just hand-jamming in values for MiniCluster.zookeepers, etc, but we could make it configurable easily enough.

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/pom.xml
```diff
@@ -126,6 +130,38 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.accumulo</groupId>
+            <artifactId>accumulo-minicluster</artifactId>
+            <version>${accumulo.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>jline</groupId>
```
jnh5y, line 160: Hmmm... you needed jline as a test dependency?  Ohhhkkk...

(Just to let you know that I still read PRs;).)

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/resources/application.conf
```diff
@@ -0,0 +1,32 @@
+geomesa = {
```
jnh5y, line 1: Hmm....  I have a suggestion that I both like and hate.

How about a common GeoMesa test jar?

On the plus side, we wouldn't have to copy this config file a zillion times across sub-modules and projects.

On the downside, we'd have one place where we could break this build without knowing it...

Thoughts?  (Since we are talking about the GeoMesa-HBase-DataStore test jar, this seems apropos.

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.geomesa.nifi.processors.AbstractGeoIngestProcessor
+import org.geomesa.nifi.processors.AbstractGeoIngestProcessor.IngestMode
+import org.geomesa.nifi.processors.datastore.PutGeoMesaAccumulo
+import org.geotools.data.DataStoreFinder
+import org.junit.{Assert, Test}
+import org.locationtech.geomesa.accumulo.MiniCluster
+import org.locationtech.geomesa.accumulo.data.AccumuloDataStoreParams
+import org.locationtech.geomesa.utils.collection.SelfClosingIterator
+
+class PutGeoMesaAccumuloTest extends LazyLogging {
```
jnh5y, line 22: First, this unit test is amazing.  One of the most useful things I've read recently!

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  @Test
+  def testIngestAttributes(): Unit = {
+    val catalog = s"${root}IngestAttributes"
+    val runner = TestRunners.newTestRunner(new PutGeoMesaAccumulo())
+    try {
+      dsParams.foreach { case (k, v) => runner.setProperty(k, v) }
+      runner.setProperty(AccumuloDataStoreParams.CatalogParam.key, catalog)
+      runner.setProperty(AbstractGeoIngestProcessor.Properties.IngestModeProp, IngestMode.Converter)
+      val attributes = new java.util.HashMap[String, String]()
```
jnh5y, line 76: Ok, to test the capability to route among SFTs/Converters, would it make sense to add a second SFT+Converter along with a test file? 

In this example, the only SFT and only Converter on the classpath are 'example' and 'example-csv'.  It isn't much of a stretch the only options end up being used;). 

(That said, if a second SFT/converter would be too much work, feel free to push back.)

Optionally, I'd be somewhat amused to see the example.csv data routed to the 'wrong' converter to verify that there is a 'failure'.  

That said, it would be more work, so stop me wherever...

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
```diff
@@ -0,0 +1,132 @@
...
+    val catalog = s"${root}IngestAttributeOverride"
+    val runner = TestRunners.newTestRunner(new PutGeoMesaAccumulo())
+    try {
+      dsParams.foreach { case (k, v) => runner.setProperty(k, v) }
+      runner.setProperty(AccumuloDataStoreParams.CatalogParam.key, catalog)
+      runner.setProperty(AbstractGeoIngestProcessor.Properties.IngestModeProp, IngestMode.Converter)
+      runner.setProperty(AbstractGeoIngestProcessor.Properties.SftNameKey, "example")
+      runner.setProperty(AbstractGeoIngestProcessor.Properties.ConverterNameKey, "example-csv")
+      val attributes = new java.util.HashMap[String, String]()
+      attributes.put(AbstractGeoIngestProcessor.Attributes.SftNameAttribute, "renamed")
```
jnh5y, line 111: Ok, I'm glad I read this unit test...  let me see if I've got this straight...

In this example, no SFT (and hence no tables) named 'renamed' existed before this FlowFile was enqueued?  But this file being ingested would have created the SFT and tables?

This seems like a Very Dangerous Feature.  One misconfiguration, and a NiFi could call for the creation of countless tables in a distributed database?  😬

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/pom.xml
elahrvivaz: minicluster won't start without it :(

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/resources/application.conf
elahrvivaz: that seems related to our other idea of building up a suite of data-store tests in index-api that could be run in each back-end. seems like it could be a good idea, and if we expose the underlying bits well enough we could plug it into nifi here too. if the tests and the data are aligned, then we wouldn't have to worry about the data changing out from under us.

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
elahrvivaz: eh, we can do that for 3.1 ;)
the processor still requires an explicit config, either through the flowfile attributes or the processor properties. since the properties aren't set here, i think it's effectively testing the case of loading them from the flowfile.

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
elahrvivaz: sounds about right :)

geomesa-nifi-processors/geomesa-nifi-processors-common/src/test/scala/org/geomesa/nifi/processors/validators/SimpleFeatureTypeValidatorTest.scala
```diff
@@ -13,7 +13,7 @@ import org.junit.Assert._
 import org.junit.Test
 import org.mockito.Mockito._
 
-class TestSimpleFeatureTypeValidator {
+class SimpleFeatureTypeValidatorTest {
```
jnh5y, line 16: This isn't supposed to run, is it?  There's no JUnit runner annotation above^^^.

pom.xml
jnh5y: Well, if you need an idea for what to get me for Christimas in July...

geomesa-nifi-processors/geomesa-nifi-processors-common/src/test/scala/org/geomesa/nifi/processors/validators/SimpleFeatureTypeValidatorTest.scala
elahrvivaz: it does run though, i verified that... maybe the annotation is only needed for specs?

jnh5y: Hmm... interesting... maybe the Surefire config (where Test.scala files are configured) takes over?

geomesa-nifi-processors/geomesa-nifi-processors-accumulo/src/test/scala/org/geomesa/nifi/processors/accumulo/PutGeoMesaAccumuloTest.scala
jnh5y: Do we have the option to *not* do it that way?  I'd hate to shred someone's database in that manner...

elahrvivaz: well, that's the whole point of this PR...

jnh5y: Is there a way to add some kind of reasonable limit here?  E.g., you can select any of the existing SFTs which are on the classpath?

Maybe the 'full' solution for what I'm wanting would be to use the Avro Schema Registry as a brokerage for SFT names?

jnh5y:
Merged as c5eb269a5d9cf04114f8cac63386b25dca7bdcab

PR #74, opened 2020-05-19
GEOMESA-2852 Separate out kafka ds processor

jeffklassen:
Signed-off-by: Jeffrey Klassen <jeffrey.s.klassen@gmail.com>


geomesa-nifi-processors/geomesa-nifi-processors-fs/pom.xml
```diff
@@ -46,33 +46,11 @@
             <artifactId>geomesa-convert-all_2.11</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>org.apache.zookeeper</groupId>
-            <artifactId>zookeeper</artifactId>
-        </dependency>
-
-        <dependency>
```
elahrvivaz, line 54: i think we need these hadoop files, as the fsds uses the hadoop api. did you try the fs nar after these changes? we probably don't need zookeeper, but let me double check that.

geomesa-nifi-processors/geomesa-nifi-processors-kafka/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-convert-all_2.11</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.zookeeper</groupId>
+            <artifactId>zookeeper</artifactId>
+        </dependency>
+
+        <dependency>
```
elahrvivaz, line 54: conversely, we probably don't need the hadoop files here for kafka.

elahrvivaz:
merged as b949e350600f54a9cbed7d268bb329a0cf05da34

PR #78, opened 2020-06-11
GEOMESA-2873 Expose flow file attributes in converters

elahrvivaz:

### DIFF
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConverterIngestProcessor.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConverterIngestProcessor.scala
@@ -59,7 +59,10 @@ trait ConverterIngestProcessor extends AbstractGeoIngestProcessor {
   }
 
   override protected def getProcessorProperties: Seq[PropertyDescriptor] =
-    Seq(converterName, ConverterSpec, ConverterErrorMode)
+    super.getProcessorProperties ++ Seq(converterName, ConverterSpec, ConverterErrorMode)
+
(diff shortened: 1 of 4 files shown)


elahrvivaz:
requires https://github.com/locationtech/geomesa/pull/2572

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConverterIngestProcessor.scala
```diff
@@ -142,7 +147,10 @@ trait ConverterIngestProcessor extends AbstractGeoIngestProcessor {
 
       val converter = converters.borrowObject()
       try {
-        val ec = converter.createEvaluationContext(EvaluationContext.inputFileParam(name))
+        val globalParams =
+          EvaluationContext.inputFileParam(name) ++
+              (if (exposeAttributes) { file.getAttributes.asScala -- Attributes.all } else { Map.empty })
```
jnh5y, line 152: Does the -- Attributes.all strip out the datastore params?  Like the password?  

If so, that's good catch.:)

jnh5y:
Merged as ec94a64220f0ca5d70933e10b6b374ad2ab9a8f0.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConverterIngestProcessor.scala
elahrvivaz: no, that just strips out the geomesa.convert and geomesa.sft configs that may be present. the datastore params aren't exposed as flow file attributes, these would be other per-file metadata (like timestamps or correlation fields)

PR #80, opened 2020-07-28
GEOMESA-2891 Use read flags to filter data store parameters

elahrvivaz:

### DIFF
--- a/geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GeoMesaKafkaProcessor.scala
+++ b/geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GeoMesaKafkaProcessor.scala
@@ -14,7 +14,9 @@ import org.apache.nifi.annotation.documentation.{CapabilityDescription, Tags}
 import org.apache.nifi.processor.ProcessContext
 import org.geomesa.nifi.datastore.processor.AbstractGeoIngestProcessor
 import org.geomesa.nifi.datastore.processor.utils.PropertyDescriptorUtils
+import org.locationtech.geomesa.kafka.data.KafkaDataStoreFactory
 import org.locationtech.geomesa.kafka.data.KafkaDataStoreFactory.KafkaDataStoreFactoryParams
+import org.locationtech.geomesa.kafka.data.KafkaDataStoreFactory.KafkaDataStoreFactoryParams.{ProducerConfig, TopicPartitions, TopicReplication}
 
 @Tags(Array("geomesa", "kafka", "streaming", "stream", "geo", "ingest", "convert", "geotools"))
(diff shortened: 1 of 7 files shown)


elahrvivaz:
depends on https://github.com/locationtech/geomesa/pull/2600

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GeoMesaKafkaProcessor.scala
```diff
@@ -27,13 +29,8 @@ abstract class GeoMesaKafkaProcessor extends AbstractGeoIngestProcessor(GeoMesaK
...
-
   // note: KafkaDataStoreFactory.ParameterInfo is consumer-oriented, but we want producer properties here
-  private val KafkaProperties = {
-    val params =
-      Seq(Brokers, Zookeepers, ZkPath, ProducerConfig, SerializationType, TopicPartitions, TopicReplication)
-    params.map(createPropertyDescriptor)
-  }
+  private val KafkaProperties =
+    createPropertyDescriptors(KafkaDataStoreFactory) ++
+        Seq(ProducerConfig, TopicPartitions, TopicReplication).map(createPropertyDescriptor)
```
jnh5y, line 35: Ok, I'm a little confused on the need to do this for ProducerConfig, TopicPartitions, and TopicReplication.  @elahrvivaz can you explain that whenever you get around to reading through GitHub messages again?

elahrvivaz: it's b/c for geoserver we only return the 'read' related props, i.e. https://github.com/locationtech/geomesa/blob/main/geomesa-kafka/geomesa-kafka-datastore/src/main/scala/org/locationtech/geomesa/kafka/data/KafkaDataStoreFactory.scala#L75
so the default descriptors that get returned don't include the producer props. maybe we should update the factory so that all the props are returned, and they can just be ignored in geoserver...

jnh5y: Ok, this is looking good enough for right now.  If you want to make some changes like that later, I'm all for it.  It seems like we are in a position where NiFi and GeoServer will now see the sensible configs for a producer and a consumer, right?

elahrvivaz: yeah I think so, consumer for geoserver and producer for nifi

jnh5y:
Merged as a34a5e52a00c926f6780d05567c7ada7dc69f549

PR #82, opened 2020-09-01
GEOMESA-2904 GeoMesa-NiFi should write GeoAvro files by attribute name

jnh5y:
Implements 'lenient' mode with unit tests.


geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -11,30 +11,94 @@ package org.geomesa.nifi.datastore.processor
...
+                  sft2: SimpleFeatureType,
+                  matchMode: String): (FeatureWriter[SimpleFeatureType, SimpleFeature], SimpleFeature) => Unit = {
+    checkCompatibleSchema(sft1, sft2) match {
+      case Success(true) =>
+        // Schemas match, so use a regular writer.
+        // TODO:  Discuss if the useProvidedFid should be true!
+        (fw: FeatureWriter[SimpleFeatureType, SimpleFeature], sf: SimpleFeature) =>
+          FeatureUtils.write(fw, sf, true)
+      case Failure(error) =>
+        if (matchMode == "lenient") {
```
elahrvivaz, line 50: would probably be good to make "lenient" and "strict" constant vals in the companion object

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -11,30 +11,94 @@ package org.geomesa.nifi.datastore.processor
...
+        } else {
+          throw error
+        }
+    }
+  }
+
+  import scala.collection.JavaConverters._
+  // Creates an adapter from one SFT to another
+  def convert(in: SimpleFeatureType, out: SimpleFeatureType): SimpleFeature => SimpleFeature = {
+    val out_geometry_ln = out.getGeometryDescriptor.getLocalName
```
elahrvivaz, line 65: [picky] usually we use camel case naming conventions for variables

elahrvivaz, line 65: could use RichSimpleFeatureType.getGeomField here - or just check for a null geometry descriptor. in most cases we do have a geometry, but i think if you added a null check then it would work regardless.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -11,30 +11,94 @@ package org.geomesa.nifi.datastore.processor
...
 import scala.util.control.NonFatal
 
+object AvroIngestProcessor {
+  val AvroMatchMode: PropertyDescriptor =
+    new PropertyDescriptor.Builder()
+      .name("Avro SFT match mode")
+      .description("Determines how Avro SFT mismatches are handled")
+      .required(false)
+      .defaultValue("strict")
+      .allowableValues("strict", "lenient")
```
elahrvivaz, line 37: instead of 'strict' and 'lenient', would it be clearer to use something like 'by attribute index' (or number/order?) and 'by attribute name'?

jnh5y: If we uses strings with spaces, will the drop down be filled in with the options?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
jnh5y: As a question....  since we are doing this in NiFi, I think we should enable this looser ingest matching in the command line tools.  (Hence https://geomesa.atlassian.net/browse/GEOMESA-2905)

If that's the case, I wonder if this function should be pushed up to GeoMesa-utils and used there?  Thoughts?

(If so, I'll bang out GM-2905 with this factored up, and we can have this PR use the function from there.)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: i think that will still work ok.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: have you tried the CLI tools? i think it should already work, since it creates a converter, and the order of attributes in a converter definition doesn't matter

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -11,30 +11,94 @@ package org.geomesa.nifi.datastore.processor
...
+      .build()
+
+  def buildWriter(sft1: SimpleFeatureType,
+                  sft2: SimpleFeatureType,
+                  matchMode: String): (FeatureWriter[SimpleFeatureType, SimpleFeature], SimpleFeature) => Unit = {
+    checkCompatibleSchema(sft1, sft2) match {
+      case Success(true) =>
+        // Schemas match, so use a regular writer.
+        // TODO:  Discuss if the useProvidedFid should be true!
```
elahrvivaz, line 46: i guess this PR actually changes the behavior from false to true? that probably makes sense to do.

jnh5y: Yes, this PR changes.  The TODO is catch our attention and make sure we are good with that decision.;)

elahrvivaz: we could always make it a property. i'm not sure if anyone is relying on the current behavior or not, but it seems like something that could mess someone up either way.

jnh5y: Indeed!  Yeah, that may be worth a property.  Let me add that in this PR.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AbstractGeoIngestProcessor.scala
```diff
@@ -303,7 +308,7 @@ object AbstractGeoIngestProcessor {
     * @param existing current feature type
     * @param input input simple feature type
     */
-  def checkCompatibleSchema(existing: SimpleFeatureType, input: SimpleFeatureType): Unit = {
+  def checkCompatibleSchema(existing: SimpleFeatureType, input: SimpleFeatureType): Try[Boolean] = {
```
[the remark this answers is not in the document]
jnh5y: Yes, lemme try that.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
jnh5y: +1

jnh5y:
> looks good! some minor comments. can you also put up a PR to update the docs on the avro processor, i.e. https://www.geomesa.org/documentation/stable/user/nifi.html#avrotoput ?

Yes!  Here you go: https://github.com/locationtech/geomesa/pull/2610

PR #83, opened 2020-09-01
GEOMESA-2906 Create GeoAvroRecordSetWriter

jnh5y:
* Implements GeoAvroRecordSetWriter.
* Uses properties or existing SimpleFeature(Type) information.

### DIFF
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
@@ -47,13 +47,13 @@ object AvroIngestProcessor {
         // Schemas match, so use a regular writer.
         // TODO:  Discuss if the useProvidedFid should be true!
         (fw: FeatureWriter[SimpleFeatureType, SimpleFeature], sf: SimpleFeature) =>
-          FeatureUtils.write(fw, sf, true)
+          FeatureUtils.write(fw, sf)
       case Failure(error) =>
         if (matchMode == LenientMatch) {
           val sfConverter = convert(sft1, sft2)
           (fw: FeatureWriter[SimpleFeatureType, SimpleFeature], sf: SimpleFeature) => {
             val sfToWrite = sfConverter(sf)
-            FeatureUtils.write(fw, sfToWrite, true)
+            FeatureUtils.write(fw, sfToWrite)
           }
         } else {
           throw error
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
@@ -1,0 +1,104 @@
+package org.geomesa.nifi.datastore.processor.records
+
+import java.io.OutputStream
+import java.util
+import java.util.Collections
+
+import org.apache.nifi.annotation.documentation.{CapabilityDescription, Tags}
+import org.apache.nifi.components.PropertyDescriptor
+import org.apache.nifi.controller.AbstractControllerService
+import org.apache.nifi.expression.ExpressionLanguageScope
+import org.apache.nifi.logging.ComponentLog
+import org.apache.nifi.processor.util.StandardValidators
+import org.apache.nifi.schema.access.SchemaNameAsAttribute
+import org.apache.nifi.serialization.record.{Record, RecordSchema}
+import org.apache.nifi.serialization.{AbstractRecordSetWriter, RecordSetWriterFactory}
+import org.geomesa.nifi.datastore.processor.records.GeoAvroRecordSetWriterFactory.{GEOMETRY_COLUMNS, TYPE_NAME}
+import org.geomesa.nifi.datastore.processor.records.GeometryEncoding.GeometryEncoding
+import org.geomesa.nifi.datastore.processor.records.SimpleFeatureRecordConverter.TypeAndEncoding
+import org.locationtech.geomesa.features.avro.AvroDataFileWriter
+import org.opengis.feature.simple.SimpleFeatureType
+
+@Tags(Array("avro", "geoavro", "result", "set", "recordset", "record", "writer", "serializer", "row"))
+@CapabilityDescription("Writes the contents of a RecordSet as GeoAvro which AvroToPutGeoMesa* Processors can use.")
+class GeoAvroRecordSetWriterFactory extends AbstractControllerService with RecordSetWriterFactory {
+  // NB: This is the same as what the InheritSchemaFromRecord Strategy does
+  override def getSchema(map: util.Map[String, String], recordSchema: RecordSchema): RecordSchema = recordSchema
+
+  override def createWriter(componentLog: ComponentLog,
+                            recordSchema: RecordSchema,
+                            outputStream: OutputStream,
+                            map: util.Map[String, String]): GeoAvroRecordSetWriter = {
+    new GeoAvroRecordSetWriter(componentLog, recordSchema, outputStream, getConfigurationContext.getProperties)
+  }
+
+  override def getSupportedPropertyDescriptors: util.List[PropertyDescriptor] =
+    Collections.singletonList(GEOMETRY_COLUMNS)
+}
+
+object GeoAvroRecordSetWriterFactory {
+  val GEOMETRY_COLUMNS: PropertyDescriptor = new PropertyDescriptor.Builder()
+    .name("Geometry Columns")
+    .description("Comma-separated list of columns with geometries with type " +
+      "(and optionally the format, (Wkt, Wkb)).  " +
+      "Example: position:Point,line:LineString.  " +
+      "Note the first field is used as a the default geometry.")
+    .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+    .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+    .required(false).build
+
+  val TYPE_NAME: PropertyDescriptor = new PropertyDescriptor.Builder()
+    .name("SimpleFeature TypeName")
+    .description("The type name for the output.")
+    .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+    .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+    .required(false).build
+}
+
+class GeoAvroRecordSetWriter(componentLog: ComponentLog, recordSchema: RecordSchema, outputStream: OutputStream, map: util.Map[PropertyDescriptor, String]) extends AbstractRecordSetWriter(outputStream) {
+  private val schemaAccessWriter = new SchemaNameAsAttribute()
+  // Wkt is assumed to be the default GeometryEncoding.
+  // TODO:  Extract this as a property?
+  private val encodings = getEncodings(map, GeometryEncoding.Wkt)
+  private val defaultGeometryColumn: Option[String] = Option(map.get(GEOMETRY_COLUMNS)).map(_.split(":")(0))
+  val typeName: Option[String] = Option(map.get(TYPE_NAME))
+  lazy val converter: SimpleFeatureRecordConverter =
+    SimpleFeatureRecordConverter.fromRecordSchema(recordSchema, encodings, GeometryEncoding.Wkt, defaultGeometryColumn, typeName)
+
+  private val sft: SimpleFeatureType = converter.sft
+  val writer = new AvroDataFileWriter(outputStream, sft)
+
+  override def writeRecord(record: Record): util.Map[String, String] = {
+    val sf = record match {
+      case sfmr: SimpleFeatureMapRecord => sfmr.sf
+      case other: Record => converter.convert(record)
+    }
+    writer.append(sf)
+    schemaAccessWriter.getAttributes(recordSchema)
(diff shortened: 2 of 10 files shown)


jnh5y:
@elahrvivaz I tossed this up as a 'draft' PR so that we can identify what are next follow-on steps.

In terms of things to resolve before we merge this PR:

1.  Deal with raw(i) vs raw(i+1) offset!
(I'm surprised that a unit test didn't break when I changed this!)
2.  Discuss if the useProvidedFid should be true!
(IMO, seems like we ought to use whatever FID is provided; it is worth discussing.)

jnh5y:
In terms of future work, I'd like to suggest:
* A property for field names to exclude
* Add tests for Wkb (since I skated over them here)
* Add property to select column as FID
* Handling for List and Map types:)
* Implement GeoAvro Reader
** Would write WKB as WKT (maybe optionally?)

Are there other ideas? 

Do any of these ideas need to be implemented in the first pass?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
```diff
@@ -34,17 +36,93 @@ object SimpleFeatureRecordConverter {
...
+        }
+        // The default geometry is set as the first column name.
+        defaultGeometryColumn.foreach(builder.setDefaultGeometry)
+        typeName.foreach(builder.setName)
+
+        builder.buildFeatureType()
+    }
+  }
+
+  private val geometryTypeMap = scala.collection.Map(
```
elahrvivaz, line 93: is this approach going to be case sensitive? do we care? (not sure how we handle that in normal sft parsing)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
```diff
@@ -87,21 +157,32 @@ object SimpleFeatureRecordConverter {
...
     }
 
     override def convert(feature: Record): SimpleFeature = {
       val raw = feature.getValues
       val values = Array.ofDim[AnyRef](converters.length)
       var i = 0
       while (i < converters.length) {
-        values(i) = converters(i).toAttribute(raw(i + 1))
+        // Why is this raw(i+1)?
+        values(i) = converters(i).toAttribute(raw(i))
```
elahrvivaz, line 169: i think this still needs to be +1 to skip over the fid, no?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
```diff
@@ -13,5 +13,11 @@ package object records {
   object GeometryEncoding extends Enumeration {
     type GeometryEncoding = Value
     val Wkt, Wkb = Value
+    def apply(encodingString: String): GeometryEncoding = {
+      encodingString match {
+        case "Wkb" => GeometryEncoding.Wkb
+        case "Wkt" => GeometryEncoding.Wkt
```
elahrvivaz, line 19: might be good to lowercase match, and add a `case _ => throw ` for clarity

pom.xml
```diff
@@ -144,11 +144,28 @@
...
             <dependency>
                 <groupId>org.apache.nifi</groupId>
                 <artifactId>nifi-schema-registry-service-api</artifactId>
                 <version>${nifi.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.nifi</groupId>
+                <artifactId>nifi-record-serialization-services</artifactId>
+                <version>${nifi.version}</version>
+                <scope>test</scope>
```
elahrvivaz, line 161: we should double check how the scope of these impacts things - nifi processors are not entirely intuitive with the expected scope of dependencies in a nar.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
jnh5y: Yes.  No, I don't think so.  Here's how we handle it: https://github.com/locationtech/geomesa/blob/main/geomesa-utils/src/main/scala/org/locationtech/geomesa/utils/geotools/sft/SimpleFeatureSpec.scala#L274-L283
;)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
jnh5y: In the unit test I wrote, it was breaking with it as +1....

pom.xml
jnh5y: Ok, thoughts about how to do that well?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,104 @@
...
+@Tags(Array("avro", "geoavro", "result", "set", "recordset", "record", "writer", "serializer", "row"))
+@CapabilityDescription("Writes the contents of a RecordSet as GeoAvro which AvroToPutGeoMesa* Processors can use.")
+class GeoAvroRecordSetWriterFactory extends AbstractControllerService with RecordSetWriterFactory {
+  // NB: This is the same as what the InheritSchemaFromRecord Strategy does
+  override def getSchema(map: util.Map[String, String], recordSchema: RecordSchema): RecordSchema = recordSchema
+
+  override def createWriter(componentLog: ComponentLog,
+                            recordSchema: RecordSchema,
+                            outputStream: OutputStream,
+                            map: util.Map[String, String]): GeoAvroRecordSetWriter = {
```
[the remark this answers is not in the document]
jnh5y: Looks like it is the FlowFile attributes?  I'd imagine that any NiFi Expressions have already been evaluated.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,104 @@
...
+}
+
+object GeoAvroRecordSetWriterFactory {
+  val GEOMETRY_COLUMNS: PropertyDescriptor = new PropertyDescriptor.Builder()
+    .name("Geometry Columns")
+    .description("Comma-separated list of columns with geometries with type " +
+      "(and optionally the format, (Wkt, Wkb)).  " +
+      "Example: position:Point,line:LineString.  " +
+      "Note the first field is used as a the default geometry.")
+    .expressionLanguageSupported(ExpressionLanguageScope.NONE)
```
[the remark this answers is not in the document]
jnh5y: Changed this one to VariableRegistry.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
jnh5y: Good idea.  Can you see if you like how I've updated it?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
elahrvivaz: i guess we're more lenient with all the other types but not geoms... eh

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
elahrvivaz: maybe the by name case is different than the by attribute case? seems odd that the unit test hasn't changed though... can you dig into the old unit test and figure out what's going on?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
elahrvivaz: looks good, thanks.

pom.xml
elahrvivaz: i was thinking just compare the dependency tree from before and after and make sure none of the existing dependencies changed scope

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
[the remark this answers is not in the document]
elahrvivaz: maybe we could rename the param to make it clearer, although if the super method has decent javadoc maybe it doesn't matter.

michaelchehreh-ccri (commented):
I had to change these around to get it to work. Tests still passed with the change, which is odd. I was hoping to debug the test but ran out of time last week. Not certain I'll have enough free time this week to do it myself

```scala
            val writer = buildWriter(reader.getSft, sft, matchMode)
```

michaelchehreh-ccri:
> I had to change [this line](https://github.com/geomesa/geomesa-nifi/blob/d749792869111e2b08b899499ec0fba66cdc706e/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala#L137) around to get it to work. Tests still passed with the change, which is odd. I was hoping to debug the test but ran out of time last week. Not certain I'll have enough free time this week to do it myself
> 
> ```scala
>             val writer = buildWriter(sft, reader.getSft, matchMode)
> ```
> 
> became
> 
> ```scala
>             val writer = buildWriter(reader.getSft, sft, matchMode)
> ```

@jnh5y Seems like the test succeeds because the order of the attributes that remain are the same, so the mappings (`0 -> 0` `1 -> 1` `4 -> 4`) are commutative. Below are the sfts
```
reader.getSft
  .getAttributeDescriptors.asScala.map(ad => (ad.getLocalName, ad.getType.getBinding.getCanonicalName))
 0 = {Tuple2@10028} "(fid,java.lang.Integer)"
 1 = {Tuple2@10029} "(name,java.lang.String)"
 2 = {Tuple2@10030} "(friends,java.lang.String)"
 3 = {Tuple2@10031} "(lastseen,java.util.Date)"
 4 = {Tuple2@10032} "(geom,org.locationtech.jts.geom.Point)"

sft
  .getAttributeDescriptors.asScala.map(ad => (ad.getLocalName, ad.getType.getBinding.getCanonicalName))
 0 = {Tuple2@10039} "(fid,java.lang.Integer)"
 1 = {Tuple2@10040} "(name,java.lang.String)"
 2 = {Tuple2@10041} "(age,java.lang.Integer)"
 3 = {Tuple2@10042} "(dtg,java.util.Date)"
 4 = {Tuple2@10043} "(geom,org.locationtech.jts.geom.Point)"
```
\
So after changing the order of the attributes in the application.conf it fails w/o changing the order in `buildWriter(...)`
```
geomesa = {
  sfts = {
    "example" = {
      attributes = [
        { name = "fid",  type = "Integer",         index = false }
        { name = "age",  type = "Integer" }
        { name = "dtg",  type = "Date",   default = true }
        { name = "geom", type = "Point",  srid = 4326, default = true }
        { name = "name", type = "String"  }
      ]
    }
  }
...
```

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -47,13 +47,13 @@ object AvroIngestProcessor {
         // Schemas match, so use a regular writer.
         // TODO:  Discuss if the useProvidedFid should be true!
         (fw: FeatureWriter[SimpleFeatureType, SimpleFeature], sf: SimpleFeature) =>
-          FeatureUtils.write(fw, sf, true)
+          FeatureUtils.write(fw, sf)
       case Failure(error) =>
         if (matchMode == LenientMatch) {
           val sfConverter = convert(sft1, sft2)
```
michaelchehreh-ccri, line 53: It's possible you could switch the order somewhere in here instead...? Didn't do the most thorough check (logically I'm not convinced it should work) but `testAvroIngestByName` ran fine with changing this to `convert(sft2, sft1)`

Also, I forgot to run **all** tests with the changes I suggested in the other thread, and turns out the changes to the sft attribute order messes up another test. Seems like I might have to let you tackle the fixing part of this :smile:

elahrvivaz:
replaced with #86

PR #86, opened 2020-10-06
GEOMESA-2906 Create GeoAvroRecordSetWriter

jnh5y:
* Implements GeoAvroRecordSetWriter.
* Uses properties or existing SimpleFeature(Type) information.

### DIFF
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
@@ -12,7 +12,7 @@ import org.apache.nifi.processor.util.StandardValidators
 import org.apache.nifi.schema.access.SchemaNameAsAttribute
 import org.apache.nifi.serialization.record.{Record, RecordSchema}
 import org.apache.nifi.serialization.{AbstractRecordSetWriter, RecordSetWriterFactory}
-import org.geomesa.nifi.datastore.processor.records.GeoAvroRecordSetWriterFactory.{GEOMETRY_COLUMNS, IncludeVisibilities, TYPE_NAME}
+import org.geomesa.nifi.datastore.processor.records.GeoAvroRecordSetWriterFactory.{GeometryColumns, VisibilitiesColumn, TypeName}
 import org.geomesa.nifi.datastore.processor.records.GeometryEncoding.GeometryEncoding
 import org.geomesa.nifi.datastore.processor.records.SimpleFeatureRecordConverter.TypeAndEncoding
 import org.locationtech.geomesa.features.avro.AvroDataFileWriter
@@ -33,22 +33,22 @@ class GeoAvroRecordSetWriterFactory extends AbstractControllerService with Recor
   }
 
   override def getSupportedPropertyDescriptors: util.List[PropertyDescriptor] =
-    Seq(GEOMETRY_COLUMNS, IncludeVisibilities).toList.asJava
+    Seq(GeometryColumns, VisibilitiesColumn).toList.asJava
 }
 
 object GeoAvroRecordSetWriterFactory {
   // TODO factor out
-  val IncludeVisibilities: PropertyDescriptor =
+  val VisibilitiesColumn: PropertyDescriptor =
     new PropertyDescriptor.Builder()
-      .name("include-visibilities")
-      .displayName("Include Visibilities")
-      .description("Include a column with visibility expressions for each row")
-      .addValidator(StandardValidators.NON_BLANK_VALIDATOR)
-      .allowableValues("false", "true")
-      .defaultValue("false")
-      .build
+      .name("visibilities-column")
+      .displayName("Visibilities Column")
+      .description("The name of a column with visibility expressions for each row.  " +
+        "If used in conjunction with the GetGeoMesaKafkaRecord processor, note that it can be configured to write out " +
+        "visibilities to a column named \"visibilities\".")
+      .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+      .required(false).build
 
-  val GEOMETRY_COLUMNS: PropertyDescriptor = new PropertyDescriptor.Builder()
+  val GeometryColumns: PropertyDescriptor = new PropertyDescriptor.Builder()
     .name("Geometry Columns")
     .description("Comma-separated list of columns with geometries with type. " +
       "Example: position:Point,line:LineString.  " +
@@ -57,7 +57,7 @@ object GeoAvroRecordSetWriterFactory {
     .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
     .required(false).build
 
-  val TYPE_NAME: PropertyDescriptor = new PropertyDescriptor.Builder()
+  val TypeName: PropertyDescriptor = new PropertyDescriptor.Builder()
     .name("SimpleFeature TypeName")
     .description("The type name for the output.")
     .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
@@ -70,19 +70,14 @@ class GeoAvroRecordSetWriter(componentLog: ComponentLog, recordSchema: RecordSch
   // Wkt is assumed to be the default GeometryEncoding.
   // TODO:  Extract this as a property?
   private val encodings = getEncodings(map, GeometryEncoding.Wkt)
-  private val defaultGeometryColumn = Option(map.get(GEOMETRY_COLUMNS)).map(_.split(":")(0))
-  private val typeName = Option(map.get(TYPE_NAME))
+  private val defaultGeometryColumn = Option(map.get(GeometryColumns)).map(_.split(":")(0))
+  private val typeName = Option(map.get(TypeName))
 
   private val geometryColumns= encodings.map { case (k, v) =>
     GeometryColumn(k, v.clazz, defaultGeometryColumn.isDefined && defaultGeometryColumn.get.equals(k))
   }.toSeq
 
-  // TODO Abstract
-  private val visField = if (java.lang.Boolean.parseBoolean(map.get(IncludeVisibilities))) {
-      Some("visibilities")
-    } else {
-      None
-    }
+  private val visField = Some(map.get(VisibilitiesColumn))
   private val recordConverterOptions = RecordConverterOptions(typeName, None, geometryColumns, visField = visField)
   private val converter = SimpleFeatureRecordConverter(recordSchema, recordConverterOptions)
 
@@ -96,7 +91,7 @@ class GeoAvroRecordSetWriter(componentLog: ComponentLog, recordSchema: RecordSch
   }
 
   private def getEncodings(descriptorToString: util.Map[PropertyDescriptor, String], defaultEncoding: GeometryEncoding): Map[String, TypeAndEncoding] = {
-    val geometryColumns = descriptorToString.get(GEOMETRY_COLUMNS)
+    val geometryColumns = descriptorToString.get(GeometryColumns)
     if (geometryColumns == null) {
       Map()
     } else {
@@ -105,7 +100,7 @@ class GeoAvroRecordSetWriter(componentLog: ComponentLog, recordSchema: RecordSch
         .map { s =>
           // TODO: Make this exception better!
           val splits = s.split(":")
-          if (splits.size < 2) throw new Exception(s"Improper configuration string: ${map.get(GEOMETRY_COLUMNS)}")
+          if (splits.size < 2) throw new Exception(s"Improper configuration string: ${map.get(GeometryColumns)}")
           val encoding = if (splits.size == 2) {
             defaultEncoding
           } else {
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
@@ -40,8 +40,7 @@ class GeoAvroRecordSetWriterFactory extends AbstractControllerService with Recor
 object GeoAvroRecordSetWriterFactory {
   val GEOMETRY_COLUMNS: PropertyDescriptor = new PropertyDescriptor.Builder()
     .name("Geometry Columns")
-    .description("Comma-separated list of columns with geometries with type " +
-      "(and optionally the format, (Wkt, Wkb)).  " +
+    .description("Comma-separated list of columns with geometries with type. " +
       "Example: position:Point,line:LineString.  " +
       "Note the first field is used as a the default geometry.")
     .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
@@ -51,7 +50,7 @@ object GeoAvroRecordSetWriterFactory {
   val TYPE_NAME: PropertyDescriptor = new PropertyDescriptor.Builder()
     .name("SimpleFeature TypeName")
     .description("The type name for the output.")
-    .expressionLanguageSupported(ExpressionLanguageScope.NONE)
+    .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
     .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
     .required(false).build
 }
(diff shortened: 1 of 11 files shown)


geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+      "Example: position:Point,line:LineString.  " +
+      "Note the first field is used as a the default geometry.")
+    .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+    .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+    .required(false).build
+
+  val TYPE_NAME: PropertyDescriptor = new PropertyDescriptor.Builder()
+    .name("SimpleFeature TypeName")
+    .description("The type name for the output.")
+    .expressionLanguageSupported(ExpressionLanguageScope.NONE)
```
elahrvivaz, line 54: any reason not to support expression language here?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  override def getSupportedPropertyDescriptors: util.List[PropertyDescriptor] =
+    Collections.singletonList(GEOMETRY_COLUMNS)
+}
+
+object GeoAvroRecordSetWriterFactory {
+  val GEOMETRY_COLUMNS: PropertyDescriptor = new PropertyDescriptor.Builder()
+    .name("Geometry Columns")
+    .description("Comma-separated list of columns with geometries with type " +
+      "(and optionally the format, (Wkt, Wkb)).  " +
```
elahrvivaz, line 44: in the record processors, i broke out geom fields config (`geom:Point,geom2:LineString`) from the encoding (wkt, wkb). that seemed easier to configure, and seems somewhat unlikely that there would actually be both wkt and wkb in the same file. any thoughts on that?

jnh5y: Yeah, I saw that.  I believe I agree with you.  It'd be nutty for someone to configure different columns with different formats.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+  private val encodings = getEncodings(map, GeometryEncoding.Wkt)
+  private val defaultGeometryColumn: Option[String] = Option(map.get(GEOMETRY_COLUMNS)).map(_.split(":")(0))
+  val typeName: Option[String] = Option(map.get(TYPE_NAME))
+
+  val geometryColumns: Seq[GeometryColumn] = encodings.map { case (k, v) =>
+    GeometryColumn(k, v.clazz, defaultGeometryColumn.isDefined && defaultGeometryColumn.get.equals(k))
+  }.toSeq
+
+  val recordConverterOptions: RecordConverterOptions = RecordConverterOptions(typeName, None, geometryColumns)
+  lazy val converter: SimpleFeatureRecordConverter =
```
elahrvivaz, line 72: making this lazy isn't getting anything since you reference it with `private val sft` below. does it need to be lazy? if so, should make all the dependent vals lazy too

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+    SimpleFeatureRecordConverter(recordSchema, recordConverterOptions)
+
+  private val sft: SimpleFeatureType = converter.sft
+  val writer = new AvroDataFileWriter(outputStream, sft)
+
+  override def writeRecord(record: Record): util.Map[String, String] = {
+    val sf = record match {
+      case sfmr: SimpleFeatureMapRecord => sfmr.sf
```
elahrvivaz, line 80: it seems like this approach might lead to inconsistencies between records coming directly from the kafka processor vs ones routed through nifi where the feature isn't present... it's a nice little hack to avoid converting the record, but i worry it might end up causing hard-to-find subtle bugs.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+      case _: Record => converter.convert(record)
+    }
+    writer.append(sf)
+    schemaAccessWriter.getAttributes(recordSchema)
+  }
+
+  private def getEncodings(descriptorToString: util.Map[PropertyDescriptor, String], defaultEncoding: GeometryEncoding): Map[String, TypeAndEncoding] = {
+    val geometryColumns = descriptorToString.get(GEOMETRY_COLUMNS)
```
elahrvivaz, line 88: if you think splitting up the wkb/wkt config is a good idea, could extract the logic from https://github.com/geomesa/geomesa-nifi/blob/main/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/RecordIngestProcessor.scala#L52 into a function and use it here as well.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+  private val sft: SimpleFeatureType = converter.sft
+  val writer = new AvroDataFileWriter(outputStream, sft)
+
+  override def writeRecord(record: Record): util.Map[String, String] = {
+    val sf = record match {
+      case sfmr: SimpleFeatureMapRecord => sfmr.sf
+      case _: Record => converter.convert(record)
+    }
+    writer.append(sf)
+    schemaAccessWriter.getAttributes(recordSchema)
```
elahrvivaz, line 84: is this always going to be the same, since it doesn't depend on the input? if so, could cache it. i admit i don't know what this method is actually supposed to return

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
jnh5y: Nope.  I'll change it to ExpressionLanguageScope.FLOWFILE_ATTRIBUTES.

elahrvivaz: you probably don't want FLOWFILE_ATTRIBUTES - otherwise you'd need to evaluate it for every incoming flowfile

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
jnh5y: https://github.com/apache/nifi/blob/main/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/AbstractRecordSetWriter.java#L58-L61
https://github.com/apache/nifi/blob/main/nifi-commons/nifi-record/src/main/java/org/apache/nifi/serialization/WriteResult.java#L42-L45

Evidently, it is a map of things to add to the FlowFile attributes...

I'm following examples like this: https://github.com/apache/nifi/blob/main/nifi-nar-bundles/nifi-standard-services/nifi-record-serialization-services-bundle/nifi-record-serialization-services/src/main/java/org/apache/nifi/csv/WriteCSVResult.java#L150

I wonder why they do not memoize the result as you are suggesting.  Thoughts?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
jnh5y: Dropping the lazy's.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
jnh5y: Thanks.  That leaves.... VARIABLE_REGISTRY!

elahrvivaz: :100:

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
elahrvivaz: yeah not sure... maybe just add some comments documenting this thread

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,122 @@
...
+  private val defaultGeometryColumn = Option(map.get(GEOMETRY_COLUMNS)).map(_.split(":")(0))
+  private val typeName = Option(map.get(TYPE_NAME))
+
+  private val geometryColumns= encodings.map { case (k, v) =>
+    GeometryColumn(k, v.clazz, defaultGeometryColumn.isDefined && defaultGeometryColumn.get.equals(k))
+  }.toSeq
+
+  // TODO Abstract
+  private val visField = if (java.lang.Boolean.parseBoolean(map.get(IncludeVisibilities))) {
+      Some("visibilities")
```
jnh5y, line 82: @elahrvivaz  Q:  How do we want to handle the visibilities field in GetGeoMesaKafkaRecord and the GeoAvroRecordSetWriterFactory?  The former has a boolean that when true leads to a field named "visibilities" being used.  Should the GeoAvroWriter do the same?  Or allow for a configurable field?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
```diff
@@ -259,6 +282,12 @@ object SimpleFeatureRecordConverter extends LazyLogging {
       case ObjectType.BOOLEAN  => new BooleanFieldConverter(name)
       case ObjectType.DATE     => new DateFieldConverter(name)
       case ObjectType.UUID     => new UuidFieldConverter(name)
+//<<<<<<< HEAD
+//      case ObjectType.GEOMETRY => GeometryToRecordField(name, encoding, encodings)
```
jnh5y, line 286: In my first version of updating this function, I ended up adding a case to handle ObjectType.GEOMETRY.  Is that still useful / necessary?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
elahrvivaz: my thought was on conversion to a record (i.e. in GetGeoMesaKafkaRecord) we can go with standardized names. On conversion from a record (i.e. in PutGeoMesa*Record) we should support a configurable name. (one can always use nifi record processors to rename and manipulate the record cols)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
elahrvivaz: i added the geometry checks outside this method since they require some other config (geom encoding, etc), i.e. [here](https://github.com/geomesa/geomesa-nifi/blob/84f08e2d9efe93030488894c461a7574ddb1bacf/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala#L160-L169) and [here](https://github.com/geomesa/geomesa-nifi/blob/84f08e2d9efe93030488894c461a7574ddb1bacf/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala#L225-L237)
I'm open to other approaches, though.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
elahrvivaz: also see https://github.com/geomesa/geomesa-nifi/pull/87 for how i handled the user data. note there will be overlap if you include user data and vis, i didn't think it mattered enough to worry about it though.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/SimpleFeatureRecordConverter.scala
jnh5y: Meh, in my first approach, I had widened one of the methods: https://github.com/geomesa/geomesa-nifi/pull/83/files#diff-694182df46af318ff33b4b7c727ff701R125 to keep the match working the same way.  

I think it is 6 of one and half of dozen of the other.  (Meaning that we can leave it the way that is.)  

If you think my first PR's approach is better, lemme know and I can refactor it in.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,118 @@
...
+    .description("The type name for the output.")
+    .expressionLanguageSupported(ExpressionLanguageScope.VARIABLE_REGISTRY)
+    .addValidator(StandardValidators.NON_EMPTY_VALIDATOR)
+    .required(false).build
+}
+
+class GeoAvroRecordSetWriter(componentLog: ComponentLog, recordSchema: RecordSchema, outputStream: OutputStream, map: util.Map[PropertyDescriptor, String]) extends AbstractRecordSetWriter(outputStream) {
+  private val schemaAccessWriter = new SchemaNameAsAttribute()
+  // Wkt is assumed to be the default GeometryEncoding.
+  // TODO:  Extract this as a property?
```
jnh5y, line 72: I'm working through using the "GeometrySerialization" property instead.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/GeoAvroRecordSetWriterFactory.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  override def createWriter(componentLog: ComponentLog,
+                            recordSchema: RecordSchema,
+                            outputStream: OutputStream,
+                            map: util.Map[String, String]): GeoAvroRecordSetWriter = {
+    new GeoAvroRecordSetWriter(componentLog, recordSchema, outputStream, getConfigurationContext.getProperties)
+  }
+
+  override def getSupportedPropertyDescriptors: util.List[PropertyDescriptor] =
+    Seq(GeometryColumns, VisibilitiesColumn, TypeName, RecordIngestProcessor.Properties.GeometrySerialization).toList.asJava
```
jnh5y, line 36: @elahrvivaz thoughts on how to best use properties which are common to the RecordIngestProcessors and the GeoAvroRecordSetWriter?

In the last push, I am trying out sharing the GeometrySerialization one.  The VisibilitiesColumn one may be a good thing to duplicate as well.  (I unintentionally created a second one of these...)

elahrvivaz: i'd probably put them in an `object Properties { ??? }` in the org.geomesa.nifi.datastore.processor.record package object

elahrvivaz (approved):
lgtm. @michaelchehreh-ccri do you have time to test?

PR #94, opened 2020-11-10
GEOMESA-2948 Fix GetKafkaRecords offset tracking

elahrvivaz:

### DIFF
--- a/geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
+++ b/geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
@@ -32,6 +32,8 @@ import org.geomesa.nifi.datastore.processor.records.Properties.GeometrySerializa
 import org.geomesa.nifi.datastore.processor.records.{GeometryEncoding, SimpleFeatureConverterOptions, SimpleFeatureRecordConverter}
 import org.geomesa.nifi.datastore.processor.utils.PropertyDescriptorUtils
 import org.geotools.data._
+import org.locationtech.geomesa.kafka.consumer.BatchConsumer.BatchResult
+import org.locationtech.geomesa.kafka.consumer.BatchConsumer.BatchResult.BatchResult
 import org.locationtech.geomesa.kafka.data.{KafkaDataStore, KafkaDataStoreParams}
 import org.locationtech.geomesa.kafka.utils.{GeoMessage, GeoMessageProcessor}
 import org.locationtech.geomesa.utils.io.{CloseWithLogging, WithClose}
@@ -64,7 +66,6 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
   private var maxBatchSize = 10000
   private var minBatchSize = 1
   private var maxLatencyMillis: Option[Long] = None
-  private var pollTimeout = 1000L
   private var factory: RecordSetWriterFactory = _
   private var consumer: Closeable = _
 
@@ -103,7 +104,7 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
       val prop = context.getProperty(RecordMaxLatency)
       Option(prop.evaluateAttributeExpressions().asTimePeriod(TimeUnit.MILLISECONDS)).map(_.longValue)
     }
-    pollTimeout = {
+    val pollTimeout = {
       val prop = context.getProperty(PollTimeout)
       prop.evaluateAttributeExpressions().asTimePeriod(TimeUnit.MILLISECONDS).longValue
     }
@@ -116,14 +117,19 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
         // override/set auto.offset.reset
         val config = KafkaDataStoreParams.ConsumerConfig.lookupOpt(base.asJava).getOrElse(new Properties())
         config.put(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, offset)
-        base + (KafkaDataStoreParams.ConsumerConfig.key -> KafkaDataStoreParams.ConsumerConfig.text(config))
+        base ++ Map(KafkaDataStoreParams.ConsumerConfig.key -> KafkaDataStoreParams.ConsumerConfig.text(config))
       }
       lazy val safeToLog = {
         val sensitive = context.getProperties.keySet().asScala.collect { case p if p.isSensitive => p.getName }
         props.map { case (k, v) => s"$k -> ${if (sensitive.contains(k)) { "***" } else { v }}" }
       }
       logger.trace(s"DataStore properties: ${safeToLog.mkString(", ")}")
-      DataStoreFinder.getDataStore(props.asJava).asInstanceOf[KafkaDataStore]
+      KafkaDataStore.LoadIntervalProperty.threadLocalValue.set(s"$pollTimeout ms")
+      try {
+        DataStoreFinder.getDataStore(props.asJava).asInstanceOf[KafkaDataStore]
+      } finally {
+        KafkaDataStore.LoadIntervalProperty.threadLocalValue.remove()
+      }
     }
     require(ds != null, "Could not load datastore using provided parameters")
 
@@ -143,7 +149,7 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
   }
 
   override def onTrigger(context: ProcessContext, session: ProcessSession): Unit = {
-    val records = processor.ready.poll(pollTimeout, TimeUnit.MILLISECONDS)
+    val records = processor.ready.poll(1, TimeUnit.SECONDS)
     if (records == null) {
       context.`yield`()
     } else {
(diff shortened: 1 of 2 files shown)


elahrvivaz:
requires https://github.com/locationtech/geomesa/pull/2647

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
```diff
@@ -139,95 +134,111 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
...
+    val done = new SynchronousQueue[Boolean]()
+
+    private var lastSuccess = System.currentTimeMillis()
+
+    override def consume(records: Seq[GeoMessage]): Boolean = {
+      val features = records.collect { case GeoMessage.Change(f) => f }
+      if (features.size < minBatchSize && maxLatencyMillis.forall(_ > System.currentTimeMillis() - lastSuccess)) {
+        logger.debug(s"Received ${features.size} records but waiting for larger batch")
+        false
+      } else if (!ready.offer(features, 10, TimeUnit.SECONDS)) {
```
jnh5y, line 234: What's this condition?  

I hate to ask, but should the 10 seconds be configurable?  (E.g., would this affect data latency for a low volume and/or bursty feed?

pom.xml
```diff
@@ -63,7 +63,7 @@
     </modules>
 
     <properties>
-        <geomesa.version>3.1.0</geomesa.version>
+        <geomesa.version>3.2.0-SNAPSHOT</geomesa.version>
```
jnh5y, line 66: Are there any JTS or Spatial4J versions that we ought to double-check?

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
```diff
@@ -139,95 +134,111 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
       val opts = SimpleFeatureConverterOptions(encoding = encoding, visField = vis, userDataField = userData)
       converter = SimpleFeatureRecordConverter(sft, opts)
       schema = factory.getSchema(Collections.emptyMap[String, String], converter.schema)
-      fs = ds.getFeatureSource(typeName)
-      fs.addFeatureListener(listener)
-    } catch {
-      case NonFatal(e) => CloseWithLogging(ds); ds = null; throw e
+      consumer = ds.createConsumer(typeName, groupId, processor)
```
jnh5y, line 137: Does a consumer require any clean up when one is done with it?

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
```diff
@@ -139,95 +134,111 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
...
+          processor.done.put(false)
         }
-      } else if (count < minBatchSize) {
-        writeRemaining(writer, timeout, count)
+      } else {
+        var ok = true
+        results.foreach { case Success(f) =>
+          try {
+            if (f.getAttribute("record.count").toInt <= 0) {
+              logger.debug("Removing flow file, no records were written")
```
jnh5y, line 192: What would happen if a flowfile writer failed/hung on a piece of data?  

Would a given batch be retried indefinitely?

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
```diff
@@ -139,95 +134,111 @@ class GetGeoMesaKafkaRecord extends AbstractProcessor {
...
-    override def changed(event: FeatureEvent): Unit = {
-      event match {
-        case e: KafkaFeatureChanged => queue.put(e.feature)
-        case _: KafkaFeatureRemoved => // no-op
-        case _: KafkaFeatureCleared => // no-op
-        case e                      => logger.error(s"Unknown event $e")
+  class RecordProcessor extends GeoMessageProcessor {
+
+    val ready = new SynchronousQueue[Seq[SimpleFeature]]()
+    val done = new SynchronousQueue[Boolean]()
```
jnh5y, line 225: Due to the use of synchronous queues (and other sensible considerations), this processor should never have more than one thread.

Is it possible in the UI to add more threads?  

If it is presently possible, can we prevent that?

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
elahrvivaz: yes, it gets closed down on line 216 when the processor is stopped

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
elahrvivaz: so errors in deserialization will be caught and logged, but won't be passed along to the client processor - see https://github.com/locationtech/geomesa/pull/2647/files#diff-db0118a973c17f07e1ee12bb8aeb0b5826e378e5c08b3dd7ab918a2aaae7ae14R63
if the client processor fails repeatedly on a piece of data for some reason, and it didn't have any logic to handle it (which this one doesn't), then yes it would flail repeatedly on the same batch. in this particular case, I'm not sure what would cause that to happen for one particular record but not others though... if you had a mis-configured recordSetWriter it seems possible, but then it seems like you would want to identify the issue and fix it rather than continuing on and dropping the data.

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
elahrvivaz: it's definitely possible in the UI. i'm not sure of any way to prevent that, although maybe there are flags or annotations you can use. i'm also not 100% sure what happens when you use multiple threads in nifi (i.e. is onTrigger just called multiple times, or is a whole new processor spun up? at least in a clustered nifi, there would presumably be different instances of the processor on different machines). this probably requires more investigation, could maybe be a follow-up ticket.

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
elahrvivaz: so this code is slightly complicated because we have 2 asynchronous callbacks that we have to coordinate - the onTrigger method and the kafka consume method. this is the amount of time that the consume method will wait for onTrigger to be invoked before bailing and giving a chance to read more messages off the kafka queue. there are 5 nifi configurations that you could tweak here - minBatchSize and maxLatency to control to the number of messages to wait for, the processor poll duration, which is how long onTrigger will wait for the consumer and defaults to 1 second, and then the nifi yield duration and poll duration, which are by default 30 seconds and 1 second, respectively. additionally, you can configure the kafka batch size and poll duration, which will control how often the consumer gets invoked. so, I could see some configuration combinations along with some kafka message time patterns that result in messages being delayed for a bit, but there are already a lot of levers to pull, so i'm not sure if making this configurable would be helpful or not.

pom.xml
elahrvivaz: haha i'm sure there are, given the recent issues we've seen... but I don't think we need to for this PR

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
elahrvivaz: the default yield is only 1 second, not 30, sorry. 30 is if the processor errors out with an uncaught exception.

geomesa-kafka-bundle/geomesa-kafka-processors/src/main/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecord.scala
jnh5y: Do we at least have a note in the docs that this processor ought to be used with just one thread?

elahrvivaz: yes it's here: https://github.com/locationtech/geomesa/blob/main/docs/user/nifi/ingress.rst#getgeomesakafkarecord

PR #101, opened 2021-01-25
GEOMESA-2992 Fix gt-processor to update classpath and validate correctly

elahrvivaz:
* Include postgis dependencies by default

### DIFF
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/pom.xml
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/pom.xml
@@ -61,6 +61,11 @@
             <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-convert-all_2.11</artifactId>
         </dependency>
+        <!-- bring in postgis to make the PutGeoTools processor work with it out of the box -->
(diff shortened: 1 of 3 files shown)


geomesa-datastore-bundle/geomesa-datastore-processors/pom.xml
```diff
@@ -61,6 +61,11 @@
             <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-convert-all_2.11</artifactId>
         </dependency>
+        <!-- bring in postgis to make the PutGeoTools processor work with it out of the box -->
+        <dependency>
+            <groupId>org.geotools.jdbc</groupId>
+            <artifactId>gt-jdbc-postgis</artifactId>
+        </dependency>
```
jnh5y, line 68: Does the bundle need to have the PostGIS JDBC driver mentioned explicitly?  Or does this work out just fine?

elahrvivaz: this seemed to work fine. i think the driver is a  transitive dependency of this artifact?

PR #102, opened 2021-03-03
GEOMESA-3016 Standardize ConvertToGeoAvro processor

elahrvivaz:
* ConvertToGeoFile processor replaces ConvertToGeoAvro
* Refactored traits to allow mixin of converter loading
* Moved mixin traits to sub packages
* Stadardize some processor properties


elahrvivaz:
depends on:
https://github.com/locationtech/geomesa/pull/2711
https://github.com/locationtech/geomesa/pull/2701
https://github.com/locationtech/geomesa/pull/2697
https://github.com/locationtech/geomesa/pull/2694

geomesa-datastore-bundle/geomesa-datastore-processors/pom.xml
```diff
@@ -66,6 +66,10 @@
             <groupId>org.geotools.jdbc</groupId>
             <artifactId>gt-jdbc-postgis</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.locationtech.geomesa</groupId>
+            <artifactId>geomesa-tools_2.11</artifactId>
```
jnh5y, line 71: Ok, why the added dependency on geomesa-tools?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
```diff
@@ -1,4 +1,5 @@
 org.geomesa.nifi.datastore.processor.ConvertToGeoAvro
+org.geomesa.nifi.datastore.processor.ConvertToGeoFile
```
jnh5y, line 2: Is this a new NiFi Processor?  If so, should there be a docs PR?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -143,41 +99,158 @@ trait AvroIngestProcessor extends FeatureTypeProcessor {
...
+          .required(false)
+          .defaultValue("false")
+          .allowableValues("true", "false")
+          .addValidator(StandardValidators.BOOLEAN_VALIDATOR)
+          .build()
+
+    @deprecated
+    val ExactMatch = "by attribute number and order"
+    @deprecated
+    val LenientMatch = "by attribute name"
```
jnh5y, line 207: Is there a replacement for this?  Or an alternative approach?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoAvro.scala
```diff
@@ -30,6 +30,8 @@ import scala.collection.JavaConverters._
 @Tags(Array("OGC", "geo", "convert", "converter", "simple feature", "geotools", "geomesa"))
 @CapabilityDescription("Convert incoming files into OGC SimpleFeature avro data files using GeoMesa Converters")
 @WritesAttribute(attribute = "mime.type", description = "the mime type of the outgoing format")
+@DeprecationNotice(alternatives = Array(classOf[ConvertToGeoFile]), reason = "Replaced with fully featured converter processor")
+@deprecated
```
jnh5y, line 34: Nice.  Should be mentioned in the upgrade guide / deprecations list.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+import org.locationtech.geomesa.tools.`export`.formats.FeatureExporter.ExportStream
+import org.locationtech.geomesa.tools.`export`.formats.FileSystemExporter.{OrcFileSystemExporter, ParquetFileSystemExporter}
+import org.locationtech.geomesa.tools.`export`.formats._
+import org.locationtech.geomesa.utils.io.{CloseWithLogging, PathUtils}
+import org.opengis.feature.simple.{SimpleFeature, SimpleFeatureType}
+
+import scala.util.control.NonFatal
+
+@Tags(Array("OGC", "geo", "convert", "converter", "simple feature", "geotools", "geomesa",
+  "parquet", "orc", "arrow", "gml", "geojson", "csv", "leaflet"))
```
jnh5y, line 39: Leaflet.  ROFL...  <sigh>

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+@CapabilityDescription("Convert incoming files into OGC data files using GeoMesa Converters")
+@WritesAttributes(
+  Array(
+    new WritesAttribute(attribute = "geomesa.convert.successes", description = "Number of features written successfully"),
+    new WritesAttribute(attribute = "geomesa.convert.failures", description = "Number of features with errors")
+  )
+)
+class ConvertToGeoFile extends ConvertInputProcessor {
```
jnh5y, line 47: There's part of me that wants to say that we ought to follow the GeoTools OGR DataStore approach and just offer up a conversion that's re-use what GDAL supports.

We'd get all these formats: https://gdal.org/drivers/vector/index.html  Surely that'd make everything happy!?!?!

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+            val exporter = format match {
+              case ExportFormat.Arrow   => new ArrowExporter(stream, ConvertToGeoFile.getArrowHints(sft), Map.empty)
+              case ExportFormat.Avro    => new AvroExporter(new FlowFileExportStream(out), gzip)
+              case ExportFormat.Bin     => new BinExporter(stream, new Hints())
+              case ExportFormat.Csv     => DelimitedExporter.csv(stream, headers)
+              case ExportFormat.Gml2    => GmlExporter.gml2(stream)
+              case ExportFormat.Gml3    => GmlExporter(stream)
+              case ExportFormat.Json    => new GeoJsonExporter(stream)
+              case ExportFormat.Leaflet => new LeafletMapExporter(stream)
+              case ExportFormat.Tsv     => DelimitedExporter.tsv(stream, headers)
```
jnh5y, line 106: These exporters are in the geomesa-tools jar, right?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+              // shouldn't happen unless someone adds a new format and doesn't implement it here
```
jnh5y, line 107: *cough*  OGR Format.  JNI to GDAL FTW!

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+          }
+        })
+      } else {
+        val dir = Files.createTempDirectory("gm-nifi")
+        try {
+          val file = new File(dir.toFile, s"export.${format.extensions.head}")
+          val exporter = format match {
+            case ExportFormat.Orc     => new OrcFileSystemExporter(file.getAbsolutePath)
+            case ExportFormat.Parquet => new ParquetFileSystemExporter(file.getAbsolutePath)
+            // case ExportFormat.Shp     => new ShapefileExporter(file) // TODO zip up dir??
```
jnh5y, line 120: Should we remove this comment and add a ticket?  Or implement?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+  )
+
+  /**
+   * Gets the hints to inform an arrow export.
+   *
+   * Currently we:
+   *   <ul>
+   *     <li>exclude fids</li>
+   *     <li>dictionary encode all string fields</li>
+   *     <li>TODO sort by date</li>
```
jnh5y, line 179: Yeah, we definitely need a ticket for this option.  I'd opine that we should do it before the 3.2.0 release.  (It ought to be easy, right?)

pom.xml
```diff
@@ -507,6 +513,51 @@
...
+                <groupId>io.netty</groupId>
+                <artifactId>netty-all</artifactId>
+                <version>${netty.version}</version>
+            </dependency>
+
+            <!-- specific to hadoop not to be confused with above-->
+            <dependency>
+                <groupId>io.netty</groupId>
+                <artifactId>netty</artifactId>
+                <version>3.6.2.Final</version>
```
jnh5y, line 558: Hmm....  everything in this jar seems to be in the org.jboss.netty package.  

From running mvn dependency:tree, it looks like there's another netty jar or two pulled in.  I don't know if we need/want to manage those versions or not.

geomesa-datastore-bundle/geomesa-datastore-processors/pom.xml
elahrvivaz: this is to pull in the exporters for the ConvertToGeoFile so we can output more than just avro!

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
elahrvivaz: yes, i need to update the docs in main gm, thanks for the reminder

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: the functionality is the same, i just standardized on the 'schema compatibility' property that was in use by the other processors. so now you'd use 'exact' and 'existing' instead of these two.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
elahrvivaz: lol, i could remove it if you want. it wasn't any extra work so i left it in

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
elahrvivaz: it supports excel so i'm on board!

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
elahrvivaz: i wasn't sure anyone would ever want this, and it was more complicated to implement since it creates multiple files... my thinking was to just leave it as a TODO in case anyone ever comes looking, since I'd rather not open a ticket we don't plan on implementing.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
elahrvivaz: mmm, i'm not sure. we don't support it in the CLI tools exporter, so we'd have to implement it there. i assume it wasn't implemented there b/c it's not trivial, or I probably would have done it at the time. again, wasn't sure if anyone would ever use this so didn't want to spend more time on it. i think it would likely be possible to pre-sort your input if you really needed to get out a sorted arrow file.

here's the block in question: https://github.com/locationtech/geomesa/blob/main/geomesa-tools/src/main/scala/org/locationtech/geomesa/tools/export/formats/ArrowExporter.scala#L57

pom.xml
elahrvivaz: added netty-codec-http, everything seems to be the same version now (except that jboss jar)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
jnh5y: Ok, that likely needs a quick note in the upgrade guide.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
jnh5y: Yeah, it was just funny.;)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
jnh5y: Le sigh, no PPT support.  Not *everyone* will be happy....  like usual.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
jnh5y: Fair.

I could see shapefile support as a way to update contextual layers.  For example, one could have a process that writes a new shapefile to S3 as the result of some process.  

I'm still on the fence about whether or not that's a great idea or not.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -11,130 +11,86 @@ package org.geomesa.nifi.datastore.processor
...
+      mode: CompatibilityMode): IngestProcessor = {
     val useProvidedFid = context.getProperty(UseProvidedFid).getValue.toBoolean
-    new AvroIngest(dataStore, writers, sftArg, typeName, matchMode, useProvidedFid)
+    val matchMode =
+      Option(context.getProperty(Properties.AvroMatchMode).getValue)
+          .collect { case m if m.nonEmpty && !ModeMappings.contains(m -> mode) =>
+            val compat = ModeMappings.collectFirst { case (old, c) if old == m => c }.get
+            logger.warn(s"Using deprecated match mode to override compatibility mode $mode with $compat")
+            compat
+          }
```
jnh5y, line 69: Can this be None?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -143,41 +99,158 @@ trait AvroIngestProcessor extends FeatureTypeProcessor {
...
+                logger.warn(
+                  s"Using deprecated property ${Properties.AvroSftName.getName} - " +
+                      s"use ${FeatureNameOverride.getName} instead")
+                old
+              } else {
+                logger.warn(
+                  s"Ignoring property ${Properties.AvroSftName.getName}: ${old.get} " +
+                      s"in favor of ${FeatureNameOverride.getName}: ${fno.get}")
+                fno
+              }
```
jnh5y, line 118: What happens if both old and fno are empty?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
```diff
@@ -143,41 +99,158 @@ trait AvroIngestProcessor extends FeatureTypeProcessor {
...
+        store.createSchema(sft)
+        None
+      } else if (SimpleFeatureTypes.compare(sft, existing) == 0) {
+        None
+      } else if (mode == CompatibilityMode.Update) {
+        logger.info(
+          s"Updating schema ${sft.getTypeName}:" +
+              s"\n  from: ${SimpleFeatureTypes.encodeType(existing)}" +
+              s"\n  to:   ${SimpleFeatureTypes.encodeType(sft)}")
+        store.updateSchema(sft.getTypeName, sft)
```
jnh5y, line 169: Wow, NiFi can call updateSchema?!  

Neat and kinda crazy/awesome.

jnh5y: UpdateSchema gets a Curator distributed lock, right?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordUpdateProcessor.scala
```diff
@@ -39,6 +39,7 @@ import scala.util.control.NonFatal
  * 'modify' mode (which require the entire feature to be input), this processor only requires specifying
  * the fields that will be updated
  */
+@SupportsBatching
```
jnh5y, line 42: We don't have any concerns about when the JavaDoc says not to use this annotation, right?

https://www.javadoc.io/doc/org.apache.nifi/nifi-api/latest/org/apache/nifi/annotation/behavior/SupportsBatching.html

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordUpdateProcessor.scala
```diff
@@ -165,15 +171,16 @@ trait RecordUpdateProcessor extends DataStoreProcessor {
             }
           })
 
-          session.putAttribute(file, Attributes.IngestSuccessCount, success.toString)
-          session.putAttribute(file, Attributes.IngestFailureCount, failure.toString)
-          session.transfer(file, SuccessRelationship)
+          var output = file
+          output = session.putAttribute(output, Attributes.IngestSuccessCount, success.toString)
+          output = session.putAttribute(output, Attributes.IngestFailureCount, failure.toString)
```
jnh5y, line 176: Ok, so I haven't read closely enough yet.  What's going on with output here?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/geotools/GeoToolsProcessor.scala
```diff
@@ -22,9 +21,9 @@ abstract class GeoToolsProcessor extends DataStoreProcessor(Seq.empty) {
 
   private var dataStoreName: PropertyDescriptor = _
 
-  override protected def init(context: ProcessorInitializationContext): Unit = {
-    dataStoreName = GeoToolsProcessor.dataStoreName(listDataStores().map(_.getDisplayName).toSeq)
-    super.init(context)
+  override def reloadDescriptors(): Unit = {
+    dataStoreName = GeoToolsProcessor.dataStoreName(GeoToolsProcessor.listDataStores().map(_.getDisplayName).toSeq)
+    super.reloadDescriptors()
```
jnh5y, line 26: Neat.  When does reloadDescriptors get called?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/mixins/BaseProcessor.scala
```diff
@@ -0,0 +1,121 @@
...
+   * @return
+   */
+  protected def getSecondaryProperties: Seq[PropertyDescriptor] = Seq(ExtraClasspaths)
+
+  /**
+   * Get tertiary processor params - these will come third in the UI
+   *
+   * @return
+   */
+  protected def getTertiaryProperties: Seq[PropertyDescriptor] = Seq.empty
```
jnh5y, line 81: This ordering bit is likely lovely for the UI!

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: i don't think so b/c it's coming from the 'allowableValues' drop down

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: that's ok, then the nameArg will be None and it will use the name from the sft in the avro file

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/AvroIngestProcessor.scala
elahrvivaz: yeah it's had that option for a while now. most stores will get a lock, not 100% sure all of them will...

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordUpdateProcessor.scala
elahrvivaz: ah, i thought it would just give you multiple flow files at once :P
this was added by Hulbert back in the original processor and we've just been carrying it forward since. i'll check how we use it, the javadocs are a bit confusing. i don't think we ever call ProcessSession.commit, maybe we should be?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordUpdateProcessor.scala
elahrvivaz: all the methods on an output flow file return a new (updated) flow file back. i wasn't seeing the attributes come through correctly without referencing the returned flow file in the converter processor, so i updated it here as well. i'm not sure if it was broken before actually? but this seems safer

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/geotools/GeoToolsProcessor.scala
elahrvivaz: onAdded is the annotation, also in the protected AbstractProcessor.init method. but seems to be called at least when the properties are updated (or maybe only properties that are marked as modifying the classpath)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/mixins/BaseProcessor.scala
elahrvivaz: yep that's why it's there

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordUpdateProcessor.scala
elahrvivaz: > All methods that are called on a ProcessSession happen as a transaction. When we decided to end the transaction, we can do so either by calling commit() or by calling rollback(). Typically, this is handled by the AbstractProcessor class: if the onTrigger method throws an Exception, the AbstractProcessor will catch the Exception, call session.rollback(), and then re-throw the Exception. Otherwise, the AbstractProcessor will call commit() on the ProcessSession.

elahrvivaz: we don't support transactions, so there's currently no way to ensure that an entire flow file is processed atomically. as such, we may end up writing or updating multiple times if for some reason a flow file fails part way through and is replayed. adding SupportsBatching would seem to make the problem "worse" in a sense, in that there could potentially be more files replayed on a failure. but it might also make the processing faster, particularly for many small files. and it seems like our ingests/updates need to be idempotent anyway. what do you think?

elahrvivaz: > It is also important to note how this behavior is affected by using the org.apache.nifi.annotation.behavior.SupportsBatching annotation. If a Processor utilizes this annotation, calls to ProcessSession.commit may not take effect immediately. Rather, these commits may be batched together in order to provide higher throughput. However, if at any point, the Processor rolls back the ProcessSession, all changes since the last call to commit will be discarded and all "batched" commits will take effect. These "batched" commits are not rolled back.

jnh5y: I'm fine continuing to carry the annotation forward.  Good notes about the annotation.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/ConvertToGeoFile.scala
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+  }
+}
+
+object ConvertToGeoFile {
+
+  import scala.collection.JavaConverters._
+
+  val Formats = Seq(
```
jnh5y, line 157: The newly supported formats also should be called out in the docs PR!

jnh5y (approved):
Approved pending a docs PR for GeoMesa!

PR #103, opened 2021-03-12
GEOMESA-3025 Configuration of write mode via flow file attribute

elahrvivaz:


elahrvivaz:
stacks on top of #102

jnh5y:
Ok, can you explain a little more about this PR?  I'm not grokking it.  Also, are there any docs changes which ought to be written up?

elahrvivaz:
this just lets you toggle between append and modify by setting a nifi expression in the field, and then adding an attribute to your flow file. so, if you want to write appends but also send through occasional modifications, you don't have to create a whole new processor. no doc change needed, since we don't currently document which attributes take nifi expression language.

jnh5y:
@elahrvivaz looks like this could use a quick rebase.

elahrvivaz:
rebased

PR #104, opened 2021-03-18
Switch to use github actions

elahrvivaz:


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,29 @@
+name: build-and-test
+
+on:
+  push:
+  pull_request:
```
jnh5y, line 5: While we are enabling GH Actions, can we set them to run nightly as well?

elahrvivaz: we can, you mean to ensure that it works with the latest snapshots?

jnh5y: Yeah, that was my goal.  I suppose we'd have to both have a nightly GH Action and make sure to increment the GM version in the pom to the nightly GM snapshot version.

Does that seem reasonable?

elahrvivaz: lets add that as a follow on task if you don't mind... we probably want to think through it.

jnh5y: Works for me.  I've entered a ticket here: https://geomesa.atlassian.net/browse/GEOMESA-3029 and tagged it for 4.0.0.

PR #105, opened 2021-03-24
GEOMESA-3031 Simplify ingest from Kafka to other backends

elahrvivaz:
* Set default nifi expressions in GeoAvroRecordSetWriter
* GetKafkaRecordProcessor fills in default expressions to eliminate manual config
* AvroToPut* processors default to use provided FID true

### DIFF
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
@@ -23,6 +23,19 @@ import org.locationtech.jts.geom.Geometry
 
 package object records {
 
+  object Expressions {
+
+    val IdCol    = new RichExpression("geomesa.id.col")
+    val GeomCols = new RichExpression("geomesa.geometry.cols")
+    val JsonCols = new RichExpression("geomesa.json.cols")
+    val DtgCol   = new RichExpression("geomesa.default.dtg.col")
+    val VisCol   = new RichExpression("geomesa.visibilities.col")
+
+    implicit class RichExpression(val attribute: String) extends AnyVal {
+      def toExpression: String = "${" + attribute + "}"
+    }
+  }
+
   object Properties {
 
     val RecordReader: PropertyDescriptor =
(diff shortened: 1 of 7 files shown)


geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/records/package.scala
```diff
@@ -323,7 +341,10 @@ package object records {
     private abstract class DynamicEvaluation[T](prop: PropertyDescriptor) extends PropertyExtractor[T] {
       override def apply(context: PropertyContext, variables: java.util.Map[String, String]): T = {
         val p = context.getProperty(prop)
-        wrap(if (p.isSet) { p.evaluateAttributeExpressions(variables).getValue } else { null })
+        val value = if (!p.isSet) { null } else {
+          Option(p.evaluateAttributeExpressions(variables).getValue).filter(_.nonEmpty).orNull
```
jnh5y, line 345: Is the change in logic here that an empty value is mapped to null?

elahrvivaz: yeah, since there's a default nifi expression in the field now, if the variable isn't set it will end up an empty string instead of null, but we don't want to actually use it.

geomesa-kafka-bundle/geomesa-kafka-processors/src/test/scala/org/geomesa/nifi/processors/kafka/GetGeoMesaKafkaRecordTest.scala
```diff
@@ -280,12 +248,40 @@ class GetGeoMesaKafkaRecordTest extends Specification with LazyLogging {
...
 
-      val bais = new ByteArrayInputStream(result)
-      val avroReader = new AvroDataFileReader(bais)
-      val featuresRead: Seq[SimpleFeature] = avroReader.toList
+      val featuresRead = WithClose(new AvroDataFileReader(new ByteArrayInputStream(result)))(_.toList)
+      featuresRead must haveSize(5)
+      featuresRead.head.visibility must beSome
+    }
+
+    "get records in GeoAvro format suitable for ingest into GeoMesa with minimal configuration" in {
```
jnh5y, line 256: Beautiful.

PR #110, opened 2021-05-12
GEOMESA-3074 Pass through original file in ConvertToGeoFile

elahrvivaz:


jnh5y:
Is a unit test worth it?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/mixins/DataStoreIngestProcessor.scala
```diff
@@ -121,7 +121,11 @@ trait DataStoreIngestProcessor extends DataStoreProcessor {
           output = session.putAttribute(output, Attributes.IngestFailureCount, result.failure.toString)
           logger.debug(
             s"Ingested file ${fullName(output)} with ${result.success} successes and ${result.failure} failures")
-          session.transfer(output, Relationships.SuccessRelationship)
+          if (result.success > 0L) {
```
jnh5y, line 124: What happens with an empty file?  Would something else handle that situation?  

Maybe that corner case isn't worth worrying about?

elahrvivaz: ah, i don't think it would be possible for the output to be empty if the success count is not zero... conversely, it is often true that the file is *not* empty even if no features were written, i.e. b/c there's avro file metadata written, or a csv header line, or whatever.

elahrvivaz:
@jnh5y ready for merge!

PR #114, opened 2021-06-17
GEOMESA-3092 Lambda data store processor

elahrvivaz:
* Nar inter-dependencies have been fixed, making most nars much smaller
* Core geomesa jars moved to geomesa-datastore-services-api-nar for inheritance
* New agnostic processors with data stores exposed via controller service
** Old data-store-specific processors are deprecated
* New Lambda data store processor

### DIFF
--- a/geomesa-accumulo-bundle/geomesa-accumulo-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
+++ b/geomesa-accumulo-bundle/geomesa-accumulo-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
@@ -2,3 +2,8 @@ org.geomesa.nifi.processors.accumulo.AvroToPutGeoMesaAccumulo
 org.geomesa.nifi.processors.accumulo.PutGeoMesaAccumulo
 org.geomesa.nifi.processors.accumulo.PutGeoMesaAccumuloRecord
 org.geomesa.nifi.processors.accumulo.UpdateGeoMesaAccumuloRecord
+org.geomesa.nifi.datastore.processor.ConvertToGeoAvro
+org.geomesa.nifi.datastore.processor.geotools.AvroToPutGeoTools
+org.geomesa.nifi.datastore.processor.geotools.PutGeoTools
+org.geomesa.nifi.datastore.processor.geotools.PutGeoToolsRecord
+org.geomesa.nifi.datastore.processor.geotools.UpdateGeoToolsRecord
--- a/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/UpdateGeoMesaRecord.scala
+++ b/geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/UpdateGeoMesaRecord.scala
@@ -1,0 +1,15 @@
+/***********************************************************************
+ * Copyright (c) 2015-2021 Commonwealth Computer Research, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Apache License, Version 2.0
+ * which accompanies this distribution and is available at
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.geomesa.nifi.datastore.processor
+
+import org.apache.nifi.annotation.documentation.{CapabilityDescription, Tags}
+
+@Tags(Array("geomesa", "geo", "update", "records", "geotools"))
+@CapabilityDescription("Update existing features in GeoMesa")
+class UpdateGeoMesaRecord extends RecordUpdateProcessor
--- a/geomesa-fs-bundle/geomesa-fs-processors/src/main/scala/org/geomesa/nifi/processors/fs/FileSystemDataStoreService.scala
+++ b/geomesa-fs-bundle/geomesa-fs-processors/src/main/scala/org/geomesa/nifi/processors/fs/FileSystemDataStoreService.scala
@@ -19,7 +19,6 @@ import org.locationtech.geomesa.fs.data.FileSystemDataStoreFactory.FileSystemDat
 @CapabilityDescription("Service for connecting to GeoMesa FileSystem stores")
 class FileSystemDataStoreService
     extends AwsDataStoreService[FileSystemDataStoreFactory](FileSystemDataStoreService.Properties, FileSystemDataStoreParams.ConfigsParam)
-// TODO partition scheme...
 
 object FileSystemDataStoreService extends PropertyDescriptorUtils {
   val Properties: Seq[PropertyDescriptor] = createPropertyDescriptors(FileSystemDataStoreFactory)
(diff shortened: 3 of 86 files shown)


elahrvivaz:
note: depends on https://github.com/locationtech/geomesa/pull/2777

geomesa-accumulo-bundle/geomesa-accumulo-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
```diff
@@ -2,3 +2,8 @@ org.geomesa.nifi.processors.accumulo.AvroToPutGeoMesaAccumulo
 org.geomesa.nifi.processors.accumulo.PutGeoMesaAccumulo
 org.geomesa.nifi.processors.accumulo.PutGeoMesaAccumuloRecord
 org.geomesa.nifi.processors.accumulo.UpdateGeoMesaAccumuloRecord
+org.geomesa.nifi.datastore.processor.ConvertToGeoAvro
+org.geomesa.nifi.datastore.processor.geotools.AvroToPutGeoTools
+org.geomesa.nifi.datastore.processor.geotools.PutGeoTools
+org.geomesa.nifi.datastore.processor.geotools.PutGeoToolsRecord
+org.geomesa.nifi.datastore.processor.geotools.UpdateGeoToolsRecord
```
jnh5y, line 9: Are these going into all the Processors?

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/UpdateGeoMesaRecord.scala
```diff
@@ -0,0 +1,15 @@
...
+ * http://www.opensource.org/licenses/apache2.0.php.
+ ***********************************************************************/
+
+package org.geomesa.nifi.datastore.processor
+
+import org.apache.nifi.annotation.documentation.{CapabilityDescription, Tags}
+
+@Tags(Array("geomesa", "geo", "update", "records", "geotools"))
+@CapabilityDescription("Update existing features in GeoMesa")
+class UpdateGeoMesaRecord extends RecordUpdateProcessor
```
jnh5y, line 15: Ok, so these "little" classes are needed since RecordUpdateProcessor can still be mixed in with another trait/interface?

(And once the deprecated classes are removed, then one can unwind this and clean it up?)

geomesa-fs-bundle/geomesa-fs-processors/src/main/scala/org/geomesa/nifi/processors/fs/FileSystemDataStoreService.scala
```diff
@@ -0,0 +1,26 @@
...
+import org.geomesa.nifi.datastore.processor.service.AwsDataStoreService
+import org.geomesa.nifi.datastore.processor.utils.PropertyDescriptorUtils
+import org.locationtech.geomesa.fs.data.FileSystemDataStoreFactory
+import org.locationtech.geomesa.fs.data.FileSystemDataStoreFactory.FileSystemDataStoreParams
+
+@Tags(Array("geomesa", "geotools", "geo", "hdfs", "s3"))
+@CapabilityDescription("Service for connecting to GeoMesa FileSystem stores")
+class FileSystemDataStoreService
+    extends AwsDataStoreService[FileSystemDataStoreFactory](FileSystemDataStoreService.Properties, FileSystemDataStoreParams.ConfigsParam)
+// TODO partition scheme...
```
jnh5y, line 22: What is the TODO here?

geomesa-kafka-bundle/geomesa-kafka-processors/pom.xml
```diff
@@ -15,13 +15,104 @@
...
             <groupId>org.geomesa.nifi</groupId>
             <artifactId>geomesa-datastore-processors</artifactId>
         </dependency>
+
         <dependency>
             <groupId>org.locationtech.geomesa</groupId>
             <artifactId>geomesa-kafka-datastore_${scala.binary.version}</artifactId>
+            <exclusions>
+                <!-- exclude dependencies provided by datastore-services-api-nar -->
+                <exclusion>
```
jnh5y, line 35: The part of me that's too quick wants to use an <exclusion>*:*</exclusion>.  I think those are deprecated, but meh.

Also, we still need some of the dependencies.... right?  That's too quick, right?

jnh5y:
Given the class loading issues that could come up, how did you test this out?

geomesa-accumulo-bundle/geomesa-accumulo-processors/src/main/resources/META-INF/services/org.apache.nifi.processor.Processor
elahrvivaz: yes, this is to keep things the same as before. previously they were defined in a META-INF file in geomesa-datastore-processors, but I removed that file so that they wouldn't be in the new modules.

geomesa-fs-bundle/geomesa-fs-processors/src/main/scala/org/geomesa/nifi/processors/fs/FileSystemDataStoreService.scala
elahrvivaz: as a convenience, there is an option to define the partition scheme in the FS processors. Since the PutGeoMesa processors are now generic, the option is not available any more. i was considering doing something here but it doesn't really make sense at the processor level since it's tied to the feature type, so i just decided not to add it. you can still define it in the feature type configuration or spec string, it's just a bit more inconvenient to re-use a feature type for the fsds... i will remove this TODO.

geomesa-kafka-bundle/geomesa-kafka-processors/pom.xml
elahrvivaz: yeah, we still need some of them. this was a PITA... i manually checked them for overlap by `unzip -l` the nars and comparing them against the dependency tree for the geomesa-datastore-services-api module. I believe having duplicates would not really matter as long as they are exactly the same (as that was the current state of things before this PR), but it cuts down on the nar size a lot.

elahrvivaz:
> Given the class loading issues that could come up, how did you test this out?

I ran through all the different back-end stores to make sure that they worked and ingested ok. More testing of the different processors is probably warranted but I think we can delay that until release time. Since nifi lets you have different versions of processors we should be able to use the older nars in a pinch if some edge case comes up (except for lambda, obviously)

geomesa-kafka-bundle/geomesa-kafka-processors/pom.xml
jnh5y: (Non-blocking question.)  Whew... is there a good way to automate this going forward? 

I suppose I'd be worried that we'll miss a dependency later.

elahrvivaz: i did not find a good way, and indeed did not get it down to zero overlapping dependencies. this is quite annoying, which is probably why i thought the nifi nar plugin would handle it (since it knows all the dependencies)

elahrvivaz: i think in most cases, it would just mean a larger nar, but wouldn't break functionality since the versions of the jars would be the same due to maven resolution. but that might not always be true, i.e. if the jar is trying to load something from geotools that is only available in the parent nar.

jnh5y: Hmm.... I wonder if the nifi nar plugin should be improved to do what we want it to do?

We can be the change we want to see in the world.

PR #116, opened 2021-07-23
GEOMESA-3108 Configurable feature type user data

elahrvivaz:


geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/mixins/FeatureTypeProcessor.scala
```diff
@@ -28,55 +28,55 @@ import org.opengis.feature.simple.SimpleFeatureType
...
-      override def load(key: SftArgs): Either[Throwable, SimpleFeatureType] =
-        SftArgResolver.getArg(key).right.map(decorate)
+      override def load(key: SftArgs): Either[Throwable, SimpleFeatureType] = SftArgResolver.getArg(key)
     }
   )
 
   override protected def getPrimaryProperties: Seq[PropertyDescriptor] = {
     sftName = FeatureTypeProcessor.Properties.sftName(SimpleFeatureTypeLoader.listTypeNames)
-    super.getPrimaryProperties ++ Seq(sftName, SftSpec, FeatureNameOverride)
+    Seq(sftName, SftSpec, FeatureNameOverride) ++ super.getPrimaryProperties
```
jnh5y, line 46: Does this change matter?  (I imagine it does; just asking to understand why.)

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordIngestProcessor.scala
```diff
@@ -79,8 +79,14 @@ trait RecordIngestProcessor extends DataStoreIngestProcessor {
         override def process(in: InputStream): Unit = {
           WithClose(recordReaderFactory.createRecordReader(file, in, logger)) { reader =>
             val converter = SimpleFeatureRecordConverter(reader.getSchema, opts)
+            val userData = loadFeatureTypeUserData(converter.sft, context, file)
+            val sft = if (userData.isEmpty) { converter.sft } else {
+              val copy = SimpleFeatureTypes.copy(converter.sft)
+              copy.getUserData.putAll(userData)
+              copy
+            }
```
jnh5y, line 87: It seems like `loadFeatureTypeUserData` could be changed to return an SFT with the userData set from the converter, context, and file. 

If that makes sense, that'd avoid repeating this block of code in two places.  A small consideration.

jnh5y: My suggestion may be tantamount to renaming the method `decorate` and widening its inputs to include the context and flowfile.

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/mixins/FeatureTypeProcessor.scala
elahrvivaz: it just changes the order in the UI... this puts the 'feature type user data' property after the feature type ones

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/RecordIngestProcessor.scala
elahrvivaz: yeah that was the original way it was done, but i thought it was maybe less prone to errors to have the feature type method not cause side effects. there are some places that we cache the feature types, and i didn't want to cache the modified ones. is that a good enough justification? i'm not too opposed to changing it if you feel strongly...

PR #125, opened 2021-12-07
GEOMESA-3159 Check to avoid adding suppressed exceptions to themselves

elahrvivaz:


elahrvivaz:
not going to back-port this

PR #128, opened 2021-12-13
GEOMESA-3160 Support multi-threading for JDBC data store ingest

elahrvivaz:
* Move FeatureWriters to separate file
* Simplify FeatureWriters class hierarchy


elahrvivaz:
Testing on some sample ais data with 4 threads, this sped up ingest from ~12 minutes to ~6 minutes

PR #134, opened 2022-10-04
Remove and ban log4j1

fdfea:


mvn.txt
```diff
@@ -0,0 +1,7263 @@
+[[1;34mINFO[m] Scanning for projects...
```
elahrvivaz, line 1: oops, think you added this by accident

PR #150, opened 2023-09-21
Update dependencies to resolve various CVEs

kylestamper:
This MR bumps the versions of vulnerable dependencies to reduce the number of critical and high CVEs from 55 down to 44.

It also parameterizes the versions of maven plugins and bumps them to more recent versions.


pom.xml
```diff
@@ -66,45 +66,53 @@
...
+        <netty.version>4.1.97.Final</netty.version>
+        <fasterxml.jackson.version>2.15.2</fasterxml.jackson.version>
         <fasterxml.woodstox.version>5.4.0</fasterxml.woodstox.version> <!-- managed from 5.3.0 -->
         <commons.net.version>3.9.0</commons.net.version> <!-- managed from 3.1 -->
         <commons.text.version>1.10.0</commons.text.version>
         <wildfly.openssl.version>1.1.3.Final</wildfly.openssl.version> <!-- managed from 1.0.7.Final -->
-        <gson.version>2.10</gson.version>
+        <gson.version>2.10.1</gson.version>
+        <snappy-java.version>1.1.10.3</snappy-java.version> <!-- managed from 1.1.1.6 -->
+        <protobuf-java.version>3.24.3</protobuf-java.version> <!-- managed from 2.5.0 -->
```
elahrvivaz, line 98: i'll verify, but i think this might cause problems with hbase

pom.xml
```diff
@@ -836,16 +903,10 @@
                 <version>3.6.2.Final</version>
             </dependency>
 
-            <!-- manage some hadoop dependencies -->
-            <dependency>
-                <groupId>commons-beanutils</groupId>
-                <artifactId>commons-beanutils</artifactId>
-                <version>1.9.4</version>
-            </dependency>
```
elahrvivaz, line 844: i would keep this here, with the comment, instead of moving it up (unless there's a reason to move it up)

kylestamper: Can do - moved it back down with the comment.

pom.xml
kylestamper: I think I have this sorted out.

pom.xml
```diff
@@ -66,45 +66,55 @@
...
+        <fasterxml.jackson.version>2.15.2</fasterxml.jackson.version>
         <fasterxml.woodstox.version>5.4.0</fasterxml.woodstox.version> <!-- managed from 5.3.0 -->
         <commons.net.version>3.9.0</commons.net.version> <!-- managed from 3.1 -->
         <commons.text.version>1.10.0</commons.text.version>
         <wildfly.openssl.version>1.1.3.Final</wildfly.openssl.version> <!-- managed from 1.0.7.Final -->
-        <gson.version>2.10</gson.version>
+        <gson.version>2.10.1</gson.version>
+        <snappy-java.version>1.1.10.3</snappy-java.version> <!-- managed from 1.1.1.6 -->
+        <protobuf-java.version>3.24.3</protobuf-java.version>
+        <protobuf-java.hbase1.version>2.6.1</protobuf-java.hbase1.version> <!-- needs to be 2.x for hbase1 -->
```
elahrvivaz, line 99: is this also getting used for hbase 2? i couldn't quite tell... if so, we could rename to just `protobuf-java.hbase.version`

geomesa-redis-bundle/pom.xml
```diff
@@ -17,4 +17,14 @@
...
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.protobuf</groupId>
+                <artifactId>protobuf-java</artifactId>
+                <version>${protobuf-java.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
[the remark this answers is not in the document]
kylestamper: Had to rearrange a bit, but it's cleaner now

pom.xml
kylestamper: It was using the same version (I wasn't sure either, actually), so I made this change.

elahrvivaz:
lgtm! I did notice one jar seemed to be downgraded in datastore-services nar: `metrics-core-3.2.6.jar` -> `metrics-core-3.2.4.jar` - do you care at all about that?

kylestamper:
> > lgtm! I did notice one jar seemed to be downgraded in datastore-services nar: `metrics-core-3.2.6.jar` -> `metrics-core-3.2.4.jar` - do you care at all about that?
> 
> Hmm, let me take a look at that real quick...

It doesn't seem to matter, I won't mess with it.

PR #151, opened 2023-10-26
Bump deps for High-severity CVEs 2023-44981 (Zookeeper) and 2023-44487 (Jetty/Netty)

kylestamper:
Tested against all GeoMesa NiFi datastores and they work properly, and the CVEs are resolved downstream when using the generated SNAPSHOT NARs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,33 +72,34 @@
         <scala.binary.version>2.12</scala.binary.version>
 
         <avro.version>1.11.3</avro.version>
-        <guava.version>32.1.2-jre</guava.version>
+        <guava.version>32.1.3-jre</guava.version>
 
         <!-- versions from geomesa -->
         <gt.version>28.2</gt.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -72,49 +72,50 @@
...
 
         <avro.version>1.11.3</avro.version>
-        <guava.version>32.1.2-jre</guava.version>
+        <guava.version>32.1.3-jre</guava.version>
 
         <!-- versions from geomesa -->
         <gt.version>28.2</gt.version>
         <accumulo.version>2.0.1</accumulo.version>
         <thrift.version>0.12.0</thrift.version>
-        <zookeeper.version>3.5.10</zookeeper.version>
```
elahrvivaz, line 81: was there any 3.5.x version that fixed the bug? i think there are some incompatibilities between these 2 versions...

kylestamper: Unfortunately no, this was the earliest version that had the fix.

PR #154, opened 2023-12-27
Minor dependency version bumps

kylestamper:
**Dependency Version Bumps**
Updated guava to `33.0.0-jre`
Updated hbase to `2.5.7-hadoop3`
Updated aws-java-sdk-core to `1.12.625`

**Maven Plugin Version Bumps**
Updated maven-compiler-plugin to `3.12.1`
Updated maven-surefire-plugin to `3.2.3`

**Testing**
All maven build tests pass, and all supported NiFi back-ends work as-expected.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         <scala.binary.version>2.12</scala.binary.version>
 
         <avro.version>1.11.3</avro.version>
-        <guava.version>32.1.3-jre</guava.version>
+        <guava.version>33.0.0-jre</guava.version>
 
         <!-- versions from geomesa -->
         <gt.version>28.2</gt.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -72,22 +72,22 @@
...
-        <hbase.version>2.5.6-hadoop3</hbase.version>
+        <hbase.version>2.5.7-hadoop3</hbase.version>
         <hbase-1.version>1.4.14</hbase-1.version>
         <kafka.version>2.8.2</kafka.version>
         <zkclient.version>0.11</zkclient.version>
         <postgres.version>42.7.0</postgres.version>
 
         <curator.version>4.3.0</curator.version>
-        <netty.version>4.1.101.Final</netty.version>
+        <netty.version>4.1.101.Final</netty.version>  <!-- NOTE: versions 4.1.102.Final+ break ConvertToGeoFileTest -->
```
elahrvivaz, line 90: we should probably figure out why and fix this...

elahrvivaz:
thanks for the updates!

pom.xml
kylestamper: Yeah, I looked into it a bit - it seems to be an issue between Apache Arrow and the most recent version(s) of Netty that others have encountered. Here's [an example](https://github.com/apache/arrow/issues/39265).

The `java.lang.NoSuchFieldError: chunkSize` error in the 2nd screenshot is the same one I was getting.

elahrvivaz: ah, so we probably will need to upgrade arrow versions... d'oh

kylestamper: That was my guess as well, but I wasn't sure what that might break downstream and was honestly too scared to find out! :grimacing:

PR #158, opened 2024-02-08
GEOMESA-3332 Update DataStoreService API for DataStore state management

fdfea:


geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/service/GeoMesaDataStoreService.scala
```diff
@@ -39,19 +39,23 @@ class GeoMesaDataStoreService[T <: DataStoreFactorySpi: ClassTag](descriptors: S
...
     if (store == null) {
-      store = tryGetDataStore(params).get
+      store = newDataStore()
     }
     store
   }
 
+  override def newDataStore(): DataStore = synchronized {
+    tryGetDataStore(params).get
+  }
```
elahrvivaz, line 49:
```suggestion
  override def newDataStore(): DataStore = tryGetDataStore(params).get
```
this doesn't need to be synchronized i don't think...

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/service/GeoMesaDataStoreService.scala
```diff
@@ -39,19 +39,23 @@ class GeoMesaDataStoreService[T <: DataStoreFactorySpi: ClassTag](descriptors: S
...
+
   override def dispose(ds: DataStore): Unit = synchronized {
     if (ds != null && !ds.eq(store)) {
       CloseWithLogging(ds)
     }
   }
 
   @OnEnabled
-  def onEnabled(context: ConfigurationContext): Unit = {
+  def onEnabled(context: ConfigurationContext): Unit = synchronized {
```
elahrvivaz, line 58: don't think this needs to be synchronized either, the nifi lifecycle ensures that it gets called in the correct order

geomesa-datastore-bundle/geomesa-datastore-services-api/src/main/java/org/geomesa/nifi/datastore/services/DataStoreService.java
```diff
@@ -13,7 +13,20 @@
 
 public interface DataStoreService extends ControllerService {
 
+    /**
+     * Get a {@code DataStore} that may be instance state for this {@code DataStoreService}.
```
elahrvivaz, line 17:
```suggestion
     * Get a {@code DataStore} that may be instance state for this {@code DataStoreService}. Clean up after use by calling {@code DataStoreService.dispose(...)} - do not dispose directly.
```

geomesa-datastore-bundle/geomesa-datastore-processors/src/main/scala/org/geomesa/nifi/datastore/processor/service/GeoMesaDataStoreService.scala
elahrvivaz: the synchronization is just needed for managing the state of the singleton `store`

elahrvivaz (approved):

PR #185, opened 2025-04-07
GEOMESA-3466 Update to GeoTools 33.0

elahrvivaz:


elahrvivaz:
I think this will pass once https://github.com/locationtech/geomesa/pull/3320 is merged and snapshots are published

